PR #7, opened 2021-03-02
Adjust Github Release Pipeline

jimson-msft:
- Publish the release version of all our components as release assets.
- Add new pipeline variables for specifying TargetCommit, previously this pipeline would use the current commit that the build was run off of.


azure-pipelines/publishing/github-release.yml
```diff
@@ -9,96 +9,95 @@ pr:
 pool: dotestlab
 
 variables:
```
shishirb-MSFT, line 11: >variables [](start = 0, length = 9)

Can we add a variable to allow us to test the pipeline without actually performing a release? Like a dry run that does everything except the last stage. Instead of the last stage, maybe we can just recursively list everything available in the artifact directory as the final step.

azure-pipelines/publishing/github-release.yml
```diff
@@ -9,96 +9,95 @@ pr:
...
+- stage: release
+  jobs:
+  - job: github_release
+    steps:
+    - task: GitHubRelease@1
+      inputs:
+        gitHubConnection: 'client2'
+        repositoryName: 'microsoft/do-client'
+        action: 'create'
+        assets: '$(Build.ArtifactStagingDirectory)/**/*-release/*'
```
shishirb-MSFT, line 96: >* [](start = 64, length = 1)

Consider specifying *.deb. That is the only asset we intend to publish, yes?

shishirb-MSFT:
For the source code archives, how was it using the do-client-0.6.0 name for the zip/gz files? Can we make it use the standard Source Code.zip/.gz filenames?

azure-pipelines/publishing/github-release.yml
```diff
@@ -9,96 +9,95 @@ pr:
...
-  - task: CmdLine@2
-    inputs:
-      script: 'rm -rf /tmp/do_publishing'
-    displayName: 'Clear publishing directory'
+stages:
+- stage: clean_build
+  jobs:
+  - job: deliveryoptimization_agent
+    steps:
+    - template: ../build/templates/doclient-lite-native-steps.yml
```
shishirb-MSFT, line 21: Building all components in the pipeline. Sweet!

azure-pipelines/publishing/github-release.yml
jimson-msft: Correct

---
In reply to: [585981245](https://github.com/microsoft/do-client/pull/7#discussion_r585981245) [](ancestors = 585981245)

azure-pipelines/publishing/github-release.yml
jimson-msft: Yeah I've added it as a parameter so the pipeline UI reflects it in a toggle checkbox

---
In reply to: [585962244](https://github.com/microsoft/do-client/pull/7#discussion_r585962244) [](ancestors = 585962244)

jimson-msft:
It just publishes whatever is in the publishing directory, so previously I had zipped the contents of the file. Anyways now we are no longer publishing a .zip file and publishing all individual packages

---
In reply to: [789292145](https://github.com/microsoft/do-client/pull/7#issuecomment-789292145) [](ancestors = 789292145)

shishirb-MSFT:
The debian packages were named as artifacts.tar.gz which was fine but I see that we are no longer archiving this. I was talking about the source code archive.

---
In reply to: [789339068](https://github.com/microsoft/do-client/pull/7#issuecomment-789339068) [](ancestors = 789339068,789292145)

azure-pipelines/publishing/github-release.yml
```diff
@@ -9,96 +9,109 @@ pr:
...
-  steps:
-  - task: CmdLine@2
-    inputs:
-      script: 'rm -rf /tmp/do_publishing'
-    displayName: 'Clear publishing directory'
+parameters:
+  - name: SkipPublishing
+    displayName: Build only (skip release stage)
+    type: boolean
+    default: false
```
shishirb-MSFT, line 20: >false [](start = 13, length = 5)

I'd keep default as true to avoid publishing accidentally. We want this pipeline to be a fully thought out execution.

azure-pipelines/publishing/github-release.yml
```diff
@@ -9,96 +9,109 @@ pr:
...
-      target: '$(Build.SourceVersion)'
-      tagSource: 'userSpecifiedTag'
-      tag: 'ms-do-client-$(release.version)'
-      title: '$(release.title) $(release.version)'
-      changeLogCompareToRelease: 'lastNonDraftReleaseByTag'
-      changeLogCompareToReleaseTag: 'ms-do-client-$(release.previous_version)'
-      changeLogType: 'commitBased'
\ No newline at end of file
+- stage: release
+  condition: and(succeeded(), eq('${{parameters.SkipPublishing}}', false))
```
shishirb-MSFT, line 97: >condition [](start = 2, length = 9)

When SkipPublishing is true, consider listing out, just to the pipeline log output, the artifacts that will be included in the release and things like tag and title. Is there a way to output the change log as well?

shishirb-MSFT:
Mind holding off on merging this until we have that meeting with Shiyi and team? I want all of us to understand how new changes are going to be consumed going forward.

azure-pipelines/publishing/github-release.yml
jimson-msft: Well, I'd expect running this pipeline at all to be with the intention of publishing. So this pipeline does two things:
Stage 1: Builds release versions of all our components for all platforms we support
Stage 2: Publishes

Do we see the use case of this pipeline being more for publishing or for validating a build of all our release pipelines? I imagine we won't be using this pipeline too regularly as a 'master' build of all our components, but rather as a 'we are ready to release'. We already have CI via our pipelines running on every check-in on the main branch.

---
In reply to: [586658460](https://github.com/microsoft/do-client/pull/7#discussion_r586658460) [](ancestors = 586658460)

jimson-msft:
Sure, I'll leave it open until then

---
In reply to: [789949112](https://github.com/microsoft/do-client/pull/7#issuecomment-789949112) [](ancestors = 789949112)

azure-pipelines/publishing/github-release.yml
jimson-msft: I see, so a true dry-run. Let me see

---
In reply to: [586662450](https://github.com/microsoft/do-client/pull/7#discussion_r586662450) [](ancestors = 586662450)

jimson-msft:
I'm not sure it's standard to rename the source code archive (taking a look at iotedge for example)

---
In reply to: [789944699](https://github.com/microsoft/do-client/pull/7#issuecomment-789944699) [](ancestors = 789944699,789339068,789292145)

azure-pipelines/publishing/github-release.yml
shishirb-MSFT: Okay, can you remove the default in that case? I'd like it to be explicitly chosen when kicking off the pipeline.

---
In reply to: [586683419](https://github.com/microsoft/do-client/pull/7#discussion_r586683419) [](ancestors = 586683419,586658460)

shishirb-MSFT:
I think they understand that snapping to a tag is the way forward. Feel free to check once signed off.

---
In reply to: [789965440](https://github.com/microsoft/do-client/pull/7#issuecomment-789965440) [](ancestors = 789965440,789949112)

azure-pipelines/publishing/github-release.yml
jimson-msft: Sure, removed

---
In reply to: [589704917](https://github.com/microsoft/do-client/pull/7#discussion_r589704917) [](ancestors = 589704917,586683419,586658460)

azure-pipelines/publishing/github-release.yml
jimson-msft: We can definitely print the artifacts to be published, release name and tag/title, but not a good way to output change log.

One hacky way of achieving the above and to print out the change log would be to publish a release, then immediately delete the release in a subsequent task if SkipPublishing is set to true.

---
In reply to: [586685709](https://github.com/microsoft/do-client/pull/7#discussion_r586685709) [](ancestors = 586685709,586662450)

shishirb-MSFT: Ok, in that case, print everything except the change log. I don't want to actually publish to github because (I think) it might trigger emails if someone is following our repo.

---
In reply to: [589764290](https://github.com/microsoft/do-client/pull/7#discussion_r589764290) [](ancestors = 589764290,586685709,586662450)

azure-pipelines/publishing/github-release.yml
```diff
@@ -9,96 +9,121 @@ pr:
...
+    steps:
+      - task: CmdLine@2
+        inputs:
+          script: |
+            echo Release title: $(Release.Title)
+            echo Release tag: $(Release.Version)
+            echo Tag used to generate changelog: $(Release.PreviousVersion)
+            echo directory to be published: $(Build.ArtifactStagingDirectory)
+            echo Content to be published:
+            echo ls -l $(Build.ArtifactStagingDirectory)
```
shishirb-MSFT, line 108: >echo [](start = 12, length = 4)

echo not needed or ls command right? We need the output from ls itself.

azure-pipelines/publishing/github-release.yml
jimson-msft: I've added a task under release to do this, and made the actual github release task the conditional one.

Do note that all that information is available anyways through the azure pipeline display -> all variables can be found as well as contents to be published, but the added task should make it explicit.

---
In reply to: [589767276](https://github.com/microsoft/do-client/pull/7#discussion_r589767276) [](ancestors = 589767276,589764290,586685709,586662450)

azure-pipelines/publishing/github-release.yml
```diff
@@ -9,96 +9,122 @@ pr:
...
+- stage: release
+  condition: succeeded()
+  jobs:
+  - job: github_release
+    condition: eq(${{parameters.SkipPublishing}}, false)
+    steps:
+    - task: DownloadBuildArtifacts@0
+      inputs:
+        buildType: 'current'
+        downloadType: 'specific'
```
shishirb-MSFT, line 104: >specific [](start = 23, length = 8)

Did you intend to specify an itemPattern to download only .deb files?

My suggestion would be to download all artifacts so that we can do things like calculate and publish hashes of each artifact (later).

jimson-msft: No, the default here is single.

The only other option is 'specific'
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-build-artifacts?view=azure-devops

I intend to download all artifacts, my guess was 'single' was not the right option to do so.
We'll see when the current build completes whether or not 'specific' downloads all artifacts.

shishirb-MSFT: I think you're right.

---
In reply to: [590816853](https://github.com/microsoft/do-client/pull/7#discussion_r590816853) [](ancestors = 590816853)

PR #10, opened 2021-03-12
Remove logging for missing json key

jimson-msft:


client-lite/src/util/do_json_parser.h
```diff
@@ -24,7 +24,6 @@ class JsonParser
         }
         catch (...)
         {
-            LOG_CAUGHT_EXCEPTION();
```
jimson-msft, line 27: This is flooding our client log files -  previously we did not log if a key was not found when using ini parser. The flooding is because based on previous pr feedback, iirc we do not want to create the json file and populate with default values. On a device without these configuration files created - the key will always be missing and the exception logging was flooding our logs.
Let calling function handle missing key and log as appropriate if needed.

shishirb-MSFT (approved):

PR #11, opened 2021-03-15
Handle Sighup

jimson-msft:


jimson-msft:
@shishirb-MSFT This is a reupload of the previous pr, unfortunately when we created the snapshot the PR did not get copied over, mind re-reviewing?

client-lite/src/config/config_manager.cpp
```diff
@@ -12,11 +12,17 @@ ConfigManager::ConfigManager() :
...
     _sdkConfigs(sdkConfigPath)
 {
 }
 
+void ConfigManager::RefreshAdminConfigs()
+{
+    _adminConfigs.Refresh();
+}
+
+
```
jimson-msft, line 25: nit: space

client-lite/src/util/do_json_parser.cpp
```diff
@@ -7,14 +7,25 @@
 std::chrono::seconds JsonParser::RefreshInterval = std::chrono::seconds(60);
 
 // Stores file path provided. Loads from the file later when a value is queried for.
-JsonParser::JsonParser(const std::string& jsonFilePath) :
+JsonParser::JsonParser(const std::string& jsonFilePath, bool alwaysCreateFile) :
     _jsonFilePath(jsonFilePath)
 {
+    if (!boost::filesystem::exists(_jsonFilePath) && alwaysCreateFile)
```
shishirb-MSFT, line 13: >alwaysCreateFile [](start = 53, length = 16)

Check alwaysCreateFile first (cheaper) in the conditional. Same comment for the force check on line 28.

client-lite/src/download/download_manager.cpp
```diff
@@ -102,12 +102,20 @@ DownloadStatus DownloadManager::GetDownloadStatus(const std::string& downloadId)
...
-    // This handles the race between shutdown and new download request coming in.
+    // This handles the race between shutdown or policy refresh and new download request coming in.
     std::unique_lock<std::shared_timed_mutex> lock(_downloadsMtx);
     _fRunning = !_downloads.empty();
     return !_fRunning;
 }
 
+void DownloadManager::RefreshConfigs() const
+{
+    _taskThread.SchedBlock([&]()
```
shishirb-MSFT, line 113: >SchedBlock [](start = 16, length = 10)

No need to block, so use Sched() with 'this' as the tag.

client-lite/src/exe/docs.cpp
```diff
@@ -31,26 +31,30 @@ class ProcessController
...
         while (true)
         {
+            /* For now, idle-shutdown mechanism is not applicable when running as a service.
+            The service will be started on boot and will be restarted automatically on failure.
+            SDK can assume docs is running and thus simplifies code for private preview. */
             if (_shutdownEvent.Wait(idleTimeout))
             {
                 break;
             }
+            if (_refreshEvent.IsSignaled())
```
shishirb-MSFT, line 49: >_refreshEvent [](start = 16, length = 13)

Combining the event signaled check with the idle loop means we still have a delay in reading config values.
Instead, accept fnRefreshConfigs in the constructor and store it as a static member and call it from _SignalHandler directly.

client-lite/src/exe/docs.cpp
```diff
@@ -62,12 +66,19 @@ class ProcessController
             DoLogInfo("Received signal %d. Initiating shutdown.", signalNumber);
             _shutdownEvent.SetEvent();
         }
+        if (signalNumber == SIGHUP)
```
shishirb-MSFT, line 69: >if [](start = 8, length = 2)

else if

client-lite/src/download/download_manager.h
```diff
@@ -27,6 +27,7 @@ class DownloadManager
     DownloadStatus GetDownloadStatus(const std::string& downloadId) const;
 
     bool IsIdle() const;
+    void RefreshConfigs() const;
```
shishirb-MSFT, line 30: >RefreshConfigs [](start = 9, length = 14)

RefreshAdminConfigs - to be clear and consistent.

client-lite/src/config/mcc_manager.h
```diff
@@ -3,8 +3,7 @@
 #include <chrono>
 #include <unordered_map>
 #include "ban_list.h"
-
-class ConfigManager;
```
shishirb-MSFT, line 7: >ConfigManager [](start = 6, length = 13)

fwd declaration should be enough here

client-lite/src/download/download_manager.cpp
jimson-msft: Should be SchedImmediate() so downloads immediately get the new configs I think

---
In reply to: [594674119](https://github.com/microsoft/do-client/pull/11#discussion_r594674119) [](ancestors = 594674119)

client-lite/src/exe/docs.cpp
```diff
@@ -16,57 +16,95 @@ namespace msdoutil = microsoft::deliveryoptimization::util::details;
...
+#include "do_noncopyable.h"
 #include "proc_launch_helper.h"
 #include "rest_http_controller.h"
 #include "rest_port_advertiser.h"
 #include "trace_sink.h"
 
 using namespace std::chrono_literals; // NOLINT(build/namespaces) how else should we use chrono literals?
 
-class ProcessController
+class ProcessController : DONonCopyable
```
jimson-msft, line 27: >: DONonCopyable [](start = 24, length = 15)

Perhaps there is no need to make this a singleton as there are no non-static members

client-lite/src/exe/docs.cpp
jimson-msft: I did some additional refactoring that might make the class more flexible if we need to support other signals, but let me know what you think

---
In reply to: [594676272](https://github.com/microsoft/do-client/pull/11#discussion_r594676272) [](ancestors = 594676272)

client-lite/src/exe/docs.cpp
```diff
@@ -16,57 +16,95 @@ namespace msdoutil = microsoft::deliveryoptimization::util::details;
...
+            _sigintHandler = std::move(fnHandler);
+        }
+        else if (signalNumber == SIGTERM)
+        {
+            _sigtermHandler = std::move(fnHandler);
+        }
+        else if (signalNumber == SIGHUP)
+        {
+            _sighupHandler = std::move(fnHandler);
+        }
```
jimson-msft, line 47: Not sure why but a switch case doesn't properly map the int signal input to the actual signal value. i.e. something like this wasn't working

switch (signal)
{
case SIGINT:
...
case SIGTERM:
...
case SIGHUP:
...
}

client-lite/src/util/do_json_parser.cpp
shishirb-MSFT: It's okay but curious why you chose to split into separate if statements rather than combine with &&.

---
In reply to: [594671740](https://github.com/microsoft/do-client/pull/11#discussion_r594671740) [](ancestors = 594671740)

client-lite/src/exe/docs.cpp
shishirb-MSFT: Non-copyable doesn't make it a singleton because you can create multiple instances of this class.

---
In reply to: [595354139](https://github.com/microsoft/do-client/pull/11#discussion_r595354139) [](ancestors = 595354139)

client-lite/src/exe/docs.cpp
shishirb-MSFT: It's strange to have a callback function callback into the caller to achieve its task. It is more straightforward to have ProcessController itself handle the shutdown in this case.
Moreover, you now have two copies of the lambda to handle SIGINT and SIGTERM.

In general, refactoring early to support a future unknown requirement is not always good because you never know whether the change as it is today will apply to that requirement.

---
In reply to: [595355095](https://github.com/microsoft/do-client/pull/11#discussion_r595355095) [](ancestors = 595355095,594676272)

client-lite/src/download/download_manager.cpp
```diff
@@ -102,12 +103,20 @@ DownloadStatus DownloadManager::GetDownloadStatus(const std::string& downloadId)
...
-    // This handles the race between shutdown and new download request coming in.
+    // This handles the race between shutdown or policy refresh and new download request coming in.
     std::unique_lock<std::shared_timed_mutex> lock(_downloadsMtx);
     _fRunning = !_downloads.empty();
     return !_fRunning;
 }
 
+void DownloadManager::RefreshAdminConfigs() const
+{
+    _taskThread.SchedImmediate([&]()
```
shishirb-MSFT, line 114: >& [](start = 32, length = 1)

Capture 'this' instead. The capture-all option is generally discouraged because it will capture more stuff than is needed. There's nothing other than 'this' to capture here but better to get used to that practice.

client-lite/src/exe/docs.cpp
shishirb-MSFT, line 47: >} [](start = 8, length = 1)

Depending on this will change based on my refactoring comment, add an else case here with DO_ASSERT(false).

client-lite/src/exe/docs.cpp
jimson-msft: Ah right, there's a convention for singleton involving static instance of the class itself and a private constructor

---
In reply to: [595411879](https://github.com/microsoft/do-client/pull/11#discussion_r595411879) [](ancestors = 595411879,595354139)

client-lite/src/exe/docs.cpp
jimson-msft: True, alright will revert

---
In reply to: [595415616](https://github.com/microsoft/do-client/pull/11#discussion_r595415616) [](ancestors = 595415616,595355095,594676272)

client-lite/src/exe/docs.cpp
jimson-msft: Think I'll revert the changes and just make the constructor accept the callback for refreshing configs

---
In reply to: [595419554](https://github.com/microsoft/do-client/pull/11#discussion_r595419554) [](ancestors = 595419554)

client-lite/src/exe/docs.cpp
```diff
@@ -26,18 +26,22 @@ using namespace std::chrono_literals; // NOLINT(build/namespaces) how else shoul
 class ProcessController
 {
 public:
-    ProcessController()
+    ProcessController(std::function<void()>& fnRefreshConfigs)
```
shishirb-MSFT, line 29: >& [](start = 43, length = 1)

Since you are using std::move, use && here

client-lite/src/exe/docs.cpp
```diff
@@ -90,7 +104,11 @@ HRESULT Run() try
 
     DoLogInfo("Started, %s", msdoutil::ComponentVersion().c_str());
 
-    ProcessController procController;
+    std::function<void()> fnRefreshConfigs = [&downloadManager]()
```
shishirb-MSFT, line 107: >    std::function<void()> [](start = 0, length = 25)

Can use auto here but consider writing the lambda inline on line 111. It will also satisfy the && parameter type.

jimson-msft:
Verified this change works on iteration 5.
Modified /etc/deliveryoptimization/admin-config.json to contain:
{
	"DOCacheHost": "www.microsoft.com"
}
ran sudo systemctl kill -s SIGHUP deliveryoptimization-agent.service  to refresh configs

With do plugin installed, did an apt install of npm and verified through logging that client configs were refreshed, bad mcc host was banned, and download finished succesfully.

client-lite/src/exe/docs.cpp
jimson-msft: Right, able to write inline due to parameter type being an rvalue reference

---
In reply to: [595575443](https://github.com/microsoft/do-client/pull/11#discussion_r595575443) [](ancestors = 595575443)

shishirb-MSFT (approved):
:shipit:

PR #19, opened 2021-04-02
More refactoring for cpp sdk for xplat support

jimson-msft:


sdk-cpp/CMakeLists.txt
```diff
@@ -58,19 +59,8 @@ if(${DO_PLATFORM_ID} EQUAL ${DO_PLATFORM_ID_LINUX})
     )
 
     add_library(${DO_SDK_LIB_NAME} SHARED "${sdk_source}")
```
shishirb-MSFT, line 61: This is the actual definition of the target. Move it outside the if block, to line 95?

shishirb-MSFT:
>         DO_PLATFORM_ID_WINDOWS=${DO_PLATFORM_ID_WINDOWS}

Consider moving the common ones outside, above line 51 and then appending platform specific ones here.

---
Refers to: sdk-cpp/CMakeLists.txt:68 in 14baa7c. [](commit_id = 14baa7c40a6f2d07f7e138aade6f70773f02fdf4, deletion_comment = False)

shishirb-MSFT:
>         DO_PLATFORM_ID_WINDOWS=${DO_PLATFORM_ID_WINDOWS}

Same goes for the sdk_source variable.

---
In reply to: [812663987](https://github.com/microsoft/do-client/pull/19#issuecomment-812663987) [](ancestors = 812663987)

---
Refers to: sdk-cpp/CMakeLists.txt:68 in 14baa7c. [](commit_id = 14baa7c40a6f2d07f7e138aade6f70773f02fdf4, deletion_comment = False)

sdk-cpp/CMakeLists.txt
```diff
@@ -79,11 +69,52 @@ if(${DO_PLATFORM_ID} EQUAL ${DO_PLATFORM_ID_LINUX})
...
+
+    set(sdk_public_includes
+        $<BUILD_INTERFACE:${${PROJECT_NAME}_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:include/${DO_SDK_LIB_NAME}>
+    )
+    set(sdk_private_includes
+        "src/internal/rest"
+        "src/internal/rest/util"
+        "src/internal"
+        ${docs_common_includes}
```
shishirb-MSFT, line 81: >docs_common_includes [](start = 10, length = 20)

${docs_common_includes} seems to be not used anymore, right? Even before this PR I mean.

sdk-cpp/include/do_config.h
```diff
@@ -27,4 +27,4 @@ void deliveryoptimization_free_version_buf(char** ppBuffer);
 }
 #endif
 
-#endif // (DO_PLATFORM_ID == DO_PLATFORM_ID_LINUX)
```
shishirb-MSFT, line 30: >(DO_PLATFORM_ID == DO_PLATFORM_ID_LINUX) [](start = 10, length = 40)

Revert?

shishirb-MSFT:
I think we will want to perform a merge commit when merging the feature branch into main, to preserve commit history, so let's make the PR titles a bit more descriptive. Like 'More sdk refactoring for cross-platform support' for this one.

sdk-cpp/CMakeLists.txt
jimson-msft: The sdk on linux is built as a .so, but is currently a static lib on linux. Prefer leaving it here until we determine that we will ship as a dll on windows

---
In reply to: [606371871](https://github.com/microsoft/do-client/pull/19#discussion_r606371871) [](ancestors = 606371871)

jimson-msft:
agreed, fixed

---
In reply to: [812666203](https://github.com/microsoft/do-client/pull/19#issuecomment-812666203) [](ancestors = 812666203)

sdk-cpp/include/do_config.h
jimson-msft: I think I prefer using // Linux or // Windows rather than the full string, meaning is conveyed and its less verbose

---
In reply to: [606373122](https://github.com/microsoft/do-client/pull/19#discussion_r606373122) [](ancestors = 606373122)

sdk-cpp/CMakeLists.txt
jimson-msft: yeah removed without issue

---
In reply to: [606372614](https://github.com/microsoft/do-client/pull/19#discussion_r606372614) [](ancestors = 606372614)

sdk-cpp/include/do_config.h
shishirb-MSFT: Ok but cpplint might complain (we'll deal with it later).

---
In reply to: [606388150](https://github.com/microsoft/do-client/pull/19#discussion_r606388150) [](ancestors = 606388150,606373122)

shishirb-MSFT (approved):
:shipit:

PR #24, opened 2021-04-13
Test refactoring for multi-platform support

jimson-msft:
- Add option to wait for keyboard input to make debugging easier


sdk-cpp/tests/dosdkcpp_testrunner.cpp
```diff
@@ -11,15 +16,30 @@ int main(int argc, char** argv)
     po::options_description desc("Options");
     desc.add_options()
         ("help,h", "Show help messages")
-        ("mcc-host,m", po::value<std::string>(), "MCC hostname to use (optional override)");
+#if (DO_PLATFORM_ID == DO_PLATFORM_ID_LINUX)
```
shishirb-MSFT, line 19: Let's keep mcc override for all platforms.

sdk-cpp/tests/dosdkcpp_testrunner.cpp
```diff
@@ -11,15 +16,26 @@ int main(int argc, char** argv)
...
         g_mccHostName = it->second.as<std::string>();
         std::cout << "Got overriden MCC host: " << g_mccHostName << '\n';
     }
 
+    auto manualStart = vm["manual-start"].as<bool>();
+    if (manualStart)
+    {
+        do
+        {
+            std::cout << '\n' << "Press a key to continue...";
```
shishirb-MSFT, line 36: You could start the test under the debugger instead of this. Just FYI.

shishirb-MSFT (approved):
:shipit:

sdk-cpp/tests/dosdkcpp_testrunner.cpp
jimson-msft: I normally start the exe and attach to process - is there a way to launch the exe and immediately attach?

---
In reply to: [612797218](https://github.com/microsoft/do-client/pull/24#discussion_r612797218) [](ancestors = 612797218)

shishirb-MSFT: Linux, use gdb ./deliveryoptimization_sdk_tests and include the --args option if you are passing args to the debuggee.
Windows, you can launch an executable from within windbg. Should be possbile with VS as well but it's been a while since I used it.

---
In reply to: [612800145](https://github.com/microsoft/do-client/pull/24#discussion_r612800145) [](ancestors = 612800145,612797218)

jimson-msft: Gotcha, Yeah I usually use the VS debugger - I'll leave as is

PR #26, opened 2021-04-14
Add an equivalent api for explicitly setting DOCacheHost

jimson-msft:
- Long-term mcc discoverability requires more thought - but this is an additional solution so as to still enable MCC while DU agent is using AIS

### DIFF
--- a/sdk-cpp/tests/config_tests.cpp
+++ b/sdk-cpp/tests/config_tests.cpp
@@ -47,6 +47,20 @@ TEST_F(ConfigTests, IoTConnectionString)
     ASSERT_EQ(value, std::string{expectedValue});
 }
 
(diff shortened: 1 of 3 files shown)


sdk-cpp/tests/config_tests.cpp
```diff
@@ -47,6 +47,20 @@ TEST_F(ConfigTests, IoTConnectionString)
     ASSERT_EQ(value, std::string{expectedValue});
 }
 
+TEST_F(ConfigTests, DOCacheHost)
+{
+    // [SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="Not credentials (false positive)")]
```
jimson-msft, line 52: Can probably remove the cred scan line here, and also add existing cred scan lines to the supression json file

PR #28, opened 2021-04-15
Add support for additional build types

jimson-msft:


shishirb-MSFT:
We'll need to rebuild cpprest on our build machines, right?

jimson-msft:
@shishirb-MSFT I know you have this task but it was low hanging fruit to work on while iterating on the com pipeline builds - I notice the agent binary size drops from 2.7MiB to 2.3MiB, no noticeable change for the SDK.

jimson-msft:
I suspect perhaps that 0.4 MiB drop is significant in that without cpprestsdk being statically linked - the binary size would actually have been reduced significantly

---
In reply to: [820759844](https://github.com/microsoft/do-client/pull/28#issuecomment-820759844) [](ancestors = 820759844)

jimson-msft:
Yep, as well as the docker images - need to add that to the pr

---
In reply to: [820759450](https://github.com/microsoft/do-client/pull/28#issuecomment-820759450) [](ancestors = 820759450)

azure-pipelines/build/templates/do-docker-jobs.yml
```diff
@@ -26,4 +26,4 @@ jobs:
     parameters:
       targetOsArch: ${{parameters.targetOS}}-${{parameters.targetArch}}
       dockerImageName: ${{parameters.dockerImageName}}
-      config: release
+      config: minsizerelease
```
shishirb-MSFT, line 29: >minsizerelease [](http://example.com/codeflow?start=14&length=14)

minsizerel

shishirb-MSFT:
Yeah, thanks for doing this.

---
In reply to: [820760902](https://github.com/microsoft/do-client/pull/28#issuecomment-820760902) [](http://example.com/codeflow?ancestors=820760902,820759844)

shishirb-MSFT (approved):
:shipit:

PR #29, opened 2021-04-16
Fix dockerfile for Deb10

jimson-msft:
Looks like the latest natively available version of gtest on Docker10 has a bug where no install targets are specified, so instead clone gtest from source

Also specify to not install build tools for the ubuntu dockerfile

### DIFF
--- a/.config/CredScanSupressions.json
+++ b/.config/CredScanSupressions.json
@@ -1,0 +1,25 @@
+{
+    "tool": "Credential Scanner",
+    "suppressions": [
+        {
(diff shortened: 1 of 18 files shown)


jimson-msft:
I've now updated all docker images to use minsizerel version of cpprestsdk, as well as our build machines themselves for the ubuntu amd64 builds. Pipelines currently running https://deliveryoptimization.visualstudio.com/client/_build

shishirb-MSFT (approved):
:shipit:

PR #34, opened 2021-05-06
Return specific errors from set_iot_connection_string

shishirb-MSFT:


sdk-cpp/src/do_config.cpp
```diff
@@ -20,23 +20,45 @@ const char* const g_doPluginAptBinName = DO_PLUGIN_APT_BIN_NAME;
...
-        boost::property_tree::write_json(filePath, configTree);
+            // No other configs are used at this time so overwrite the whole file here
+            boost::property_tree::write_json(filePath.string(), configTree);
 
-        return 0;
+            return 0;
+        }
+        catch (const boost::property_tree::ptree_bad_data&)
+        {
+            returnValue = -1;
```
jimson-msft, line 42: Would it be useful to have an enum representing these different errors?

jimson-msft (approved):

sdk-cpp/src/do_config.cpp
shishirb-MSFT: Maybe in future when we are writing more configs from the sdk.

PR #35, opened 2021-05-06
Return specific errors from creating and accessing the destination file path

shishirb-MSFT:


client-lite/src/util/do_file.cpp
```diff
@@ -0,0 +1,57 @@
+#include "do_common.h"
+#include "do_file.h"
+
+#include <fcntl.h>
+#include <sys/stat.h>
```
jimson-msft, line 5: Cool, the usage of posix apis potentially allows us to cut a little bloat (boost::filesystem) from our components

client-lite/src/util/do_file.cpp
```diff
@@ -0,0 +1,57 @@
...
+DOFile DOFile::Open(const std::string& path)
+{
+    int fd = open(path.data(), O_APPEND | O_WRONLY);
+    if (fd == -1)
+    {
+        THROW_HR_MSG(HRESULT_FROM_XPLAT_SYSERR(errno), "Cannot open file at %s", path.data());
+    }
+
+    return DOFile{fd};
+}
```
jimson-msft, line 35: Thanks for handling the task of returning more specific error codes

jimson-msft (approved):
:shipit:

PR #36, opened 2021-05-06
Cleanup download in all cases, avoid overwriting existing files

shishirb-MSFT:


sdk-cpp/src/do_download.cpp
```diff
@@ -58,23 +58,61 @@ void download::download_url_to_path(const std::string& uri, const std::string& d
...
 void download::download_url_to_path(const std::string& uri, const std::string& downloadFilePath, const std::atomic_bool& isCancelled, std::chrono::seconds timeOut)
 {
     constexpr std::chrono::seconds maxPollTime = 5s;
     std::chrono::milliseconds pollTime = 500ms;
     const auto endTime = std::chrono::system_clock::now() + timeOut;
 
     download oneShotDownload(uri, downloadFilePath);
+    download_cleanup_call scopedCleanup{&oneShotDownload};
+
     oneShotDownload.start();
```
shishirb-MSFT, line 107: I noticed that a failure in start() was causing abort() on line 133 to not get called (exception gets thrown), leaving a stale download object in the agent.

sdk-cpp/src/do_download.cpp
```diff
@@ -58,23 +58,61 @@ void download::download_url_to_path(const std::string& uri, const std::string& d
     download_url_to_path(uri, downloadFilePath, isCancelled, timeOut);
 }
 
+// Ensures download is finalized/aborted in the agent to avoid stale download objects.
+// Future: We could expect callers to explicitly opt-in to persist downloads in the agent and
+//   use that as an indicator to perform/skip the cleanup directly in ~download().
+class download_cleanup_call
```
jimson-msft, line 64: Does adding this cleanup impact the dosvc client? If so, consider adding this logic to the specific download_impl class

client-lite/src/download/download.cpp
```diff
@@ -311,9 +311,11 @@ void Download::_Abort() try
     _timer.Stop();
     _fileStream.Close();
     _CancelTasks();
-    if (!_destFilePath.empty())
+    // Delete file only if this download is the creator/owner. The abort could be from an
+    // error downloading to an already existing file.
+    if (_fDestFileCreated && !_destFilePath.empty())
     {
-        boost::filesystem::remove(_destFilePath);
```
jimson-msft, line 316: >        boost::filesystem::remove(_destFilePath) [](http://example.com/codeflow?start=0&length=48)

Unrelated to this change but part of cleaning up our dependencies, the change you introduced already helps us remove usage of boost::filesystem, so there's potential to see if we can just remove some of the boost libraries in enterity

sdk-cpp/src/do_download.cpp
```diff
@@ -58,23 +58,61 @@ void download::download_url_to_path(const std::string& uri, const std::string& d
     download_url_to_path(uri, downloadFilePath, isCancelled, timeOut);
 }
 
+// Ensures download is finalized/aborted in the agent to avoid stale download objects.
+// Future: We could expect callers to explicitly opt-in to persist downloads in the agent and
+//   use that as an indicator to perform/skip the cleanup directly in ~download().
```
jimson-msft, line 63: Does the agent currently have a self-cleanup mechanism to abort/expire jobs after a set period of time (24hr?)

sdk-cpp/tests/download_tests.cpp
```diff
@@ -146,32 +164,37 @@ TEST_F(DownloadTests, SimpleDownloadTest_With404UrlAndMalformedPath)
...
+    }
+    catch (const msdo::exception& e)
+    {
+       ASSERT_EQ(e.error_code(), DO_ERROR_FROM_XPLAT_SYSERR(EEXIST));
+    }
+    simpleDownload2.abort();
+    ASSERT_TRUE(boost::filesystem::exists(g_tmpFileName)); // not deleted, the earlier download is still active
+
+    simpleDownload.abort();
+    ASSERT_FALSE(boost::filesystem::exists(g_tmpFileName));
```
jimson-msft, line 196: what happens in the following case:

download 1 created
download 2 created
download 1 started
download 1 paused
download 1 abort
download 2 start

download 2 should expectedly succeed in this case correct?

jimson-msft (dismissed):
:shipit:

sdk-cpp/src/do_download.cpp
shishirb-MSFT: Windows dosvc does but the expiration is 30 days. ADU DO keeps it until shutdown (no persistence).

sdk-cpp/src/do_download.cpp
shishirb-MSFT: This is required behavior for both windows and xplat.

client-lite/src/download/download.cpp
shishirb-MSFT: boost::filesystem is still used in other parts like the rest_port_advetiser and finder. The biggest boost library is boost.log so I'm thinking of looking into that after the cpprestsdk work item.

sdk-cpp/tests/download_tests.cpp
shishirb-MSFT: yes, I'll update the the test with this.

sdk-cpp/src/do_download.cpp
jimson-msft: And it's stored in memory until then? This could explain the memory usage I saw earlier (where after many downloads, our memory usage increased) - though this would only occur in the case the start threw. so maybe not

jimson-msft (approved):
:shipit:

PR #37, opened 2021-05-12
SDK: Use Boost.Beast instead of cpprestsdk

shishirb-MSFT:


shishirb-MSFT:
Installing 1.66 version of boost.asio on top of 1.65 version for all other sub-libraries causes issues. Looking into a different lib now.

PR #41, opened 2021-06-12
Strip unneeded symbols from minsizerel binaries

shishirb-MSFT:


build/bootstrap/bootstrap-debian-10.sh
```diff
@@ -31,15 +35,15 @@ ninja
...
-mkdir cmake 
+mkdir cmake
 cd cmake
 cmake /tmp/gtest
-make 
-make install 
+make
+make install
 
 echo "Finished bootstrapping"
```
jimson-msft, line 49: Something I've been thinking about based on other boostrap scripts is to have all of these consolidated into a single bootstrap script, where platform is automatically detected. That way we can deduplicate some of the script commands to install things like python, make, etc.

I think based on what I've seen, it's also fairly standard to have the bootstrap scripts do a build of whatever the repo is before completing to verify things work correctly.

Totally unrelated to this PR though, I'll open a new task for the above:
[https://microsoft.visualstudio.com/OS/_workitems/edit/32499146/?workitem=33759607](https://microsoft.visualstudio.com/OS/_workitems/edit/32499146/?workitem=33759607)

build/bootstrap/bootstrap-debian-10.sh
```diff
@@ -5,13 +5,17 @@ set -e
...
 apt-get update -y --fix-missing
 apt-get install -y make build-essential g++ gdb gdbserver gcc git wget
 apt-get install -y python3 cmake ninja-build
 
 # Open-source library dependencies
-apt-get install -y libboost-all-dev libgtest-dev libproxy-dev libmsgsl-dev libssl-dev uuid-dev
+# Boost libs for DO
+apt-get install -y libboost-system-dev libboost-log-dev libboost-filesystem-dev libboost-program-options-dev
+# Additional Boost libs for cpprestsdk
+apt-get install -y libboost-random-dev libboost-regex-dev
```
jimson-msft, line 17: We're still pending the task to remove cpprest for the http server implementation within the agent correct?

jimson-msft:
> 

I use vscode to debug via attach to process, I imagine stripping symbols from the exe may require vscode to be provided the .pdb path now. No worries on this - just thinking out loud.

---
Refers to: common/cmake/do-build-helpers.cmake:138 in 284d3fa. [](commit_id = 284d3fa4671f907a57d9a597b9f5d06140c7de41, deletion_comment = False)

common/cmake/do-build-helpers.cmake
```diff
@@ -104,3 +104,34 @@ macro (add_do_version_lib target_name maj_min_patch_ver)
...
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    # Ensure that objcopy is present
+    find_program(OBJCOPY objcopy)
+    if (OBJCOPY STREQUAL "OBJCOPY-NOTFOUND")
+        message(FATAL_ERROR "objcopy not found")
+    endif()
+endif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+
+# From https://github.com/dotnet/coreclr/pull/3872/files
+# For minsizerel build, split unneeded symbols from target binary file and into a separate .dbg file.
```
jimson-msft, line 117: >minsizerel [](http://example.com/codeflow?start=6&length=10)

Why are symbols present in the minsizerel binary? There's a seperate CMAKE_BUILD_TYPE, RelWithDebInfo, I would have assumed both Release and MinSizeRel would have automatically excluded symbols

jimson-msft (approved):
:shipit:

build/bootstrap/bootstrap-debian-10.sh
shishirb-MSFT: Yes, I'm leaving that for Nickel.

build/bootstrap/bootstrap-debian-10.sh
shishirb-MSFT: Yes, de-duplicate and a test build sounds great.

common/cmake/do-build-helpers.cmake
shishirb-MSFT: Don't know. That's how gcc works. I should mention that it is stripping more than just debug symbols. I didn't dig to find exactly what is stripped (will do that when there's time).

shishirb-MSFT:
> 

vscode uses gdb and all I had to do for gdb is to ensure that the .dbg file was in the same directory as the binary.

---
In reply to: [860969043](https://github.com/microsoft/do-client/pull/41#issuecomment-860969043) [](http://example.com/codeflow?ancestors=860969043,860969043)

---
Refers to: common/cmake/do-build-helpers.cmake:138 in 284d3fa. [](commit_id = 284d3fa4671f907a57d9a597b9f5d06140c7de41, deletion_comment = False)

PR #43, opened 2021-06-16
Agent: Move log location to /var/log

shishirb-MSFT:


client-lite/build/postinst.in.sh
```diff
@@ -43,11 +43,12 @@ configure_dir() {
     fi
     chgrp -R $do_group_name $dir_path
     chown $do_user_name $dir_path
-    chmod g+w $dir_path
 }
 
 configure_dir "$config_path"
-configure_dir "$persistence_path"
+# group needs write permission to config path (SDK writing configs)
+chmod g+w $config_path
```
jimson-msft, line 50: possible to still have this within configure_dir()? where configure_dir() accepts a bool param to check whether it should toggle group write

sdk-cpp/tests/download_tests.cpp
```diff
@@ -532,19 +532,22 @@ TEST_F(DownloadTests, SimpleBlockingDownloadTest_ClientNotRunningPortFilePresent
...
+    auto startService = dotest::util::scope_exit([]()
+        {
+            TestHelpers::StartService("deliveryoptimization-agent.service");
+        });
     TestHelpers::CreateRestPortFiles(5);
+    ASSERT_GE(TestHelpers::CountRestPortFiles(), 5u);
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+    startService.reset();
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+    ASSERT_EQ(TestHelpers::CountRestPortFiles(), 1u) << "All other restport files must be deleted by the agent";
```
jimson-msft, line 547: Could also check to see if the portfile created has a newer timestamp than the prexisting ones, to ensure that the file created was by the daemon upon restart.

jimson-msft (approved):
:shipit:

client-lite/build/postinst.in.sh
shishirb-MSFT: Not worth it I felt.

sdk-cpp/tests/download_tests.cpp
shishirb-MSFT: If it wasn't created by the daemon, the download will fail right?

sdk-cpp/tests/download_tests.cpp
```diff
@@ -532,19 +532,22 @@ TEST_F(DownloadTests, SimpleBlockingDownloadTest_ClientNotRunningPortFilePresent
...
-// This will ensure that the actual rest port file will have the latest timestamp.
-#if 0
+    TestHelpers::StopService("deliveryoptimization-agent.service");
+    auto startService = dotest::util::scope_exit([]()
+        {
+            TestHelpers::StartService("deliveryoptimization-agent.service");
+        });
     TestHelpers::CreateRestPortFiles(5);
+    ASSERT_GE(TestHelpers::CountRestPortFiles(), 5u);
+    std::this_thread::sleep_for(std::chrono::seconds(2));
```
shishirb-MSFT, line 544: >std::this_thread::sleep_for(std::chrono::seconds(2)); [](http://example.com/codeflow?start=4&length=53)

I didn't mean to retain this sleep (was debugging). Will remove.

sdk-cpp/tests/download_tests.cpp
jimson-msft: true

jimson-msft (approved):

PR #45, opened 2021-06-26
Agent: Update MCC fallback logic

shishirb-MSFT:


client-lite/src/config/mcc_manager.h
```diff
@@ -11,14 +11,26 @@ class MCCManager
...
+    public:
+        MccHost(const std::string& address);
+        void Ban(std::chrono::seconds banInterval);
+
+        const std::string& Address() const noexcept { return address; }
+        bool IsBanned() const;
+
+    private:
+        std::string address;
+        std::chrono::steady_clock::time_point timeOfUnban;
```
jimson-msft, line 31: nit: private members so _ prefix to variable name

client-lite/src/config/mcc_manager.h
```diff
@@ -11,14 +11,26 @@ class MCCManager
...
-    bool NoFallback() const;
-    bool ReportHostError(HRESULT hr, const std::string& host);
+    boost::optional<std::chrono::seconds> FallbackDelay();
+    std::string GetHost(const std::string& originalUrl);
+    void ReportHostError(HRESULT hr, UINT httpStatusCode, const std::string& mccHost, const std::string& originalUrl);
+    bool IsBanned(const std::string& mccHost, const std::string& originalUrl) const;
 
 private:
-    bool _IsFallbackDue(const std::string& host) const;
+    struct MccHost
```
jimson-msft, line 20: Generally, I see struct's being used as passive data structures ([https://en.wikipedia.org/wiki/Passive_data_structure](https://en.wikipedia.org/wiki/Passive_data_structure))

Since this contains features of OOP (access modifiers, member functions) - would it be better to have this as a class?

client-lite/src/config/mcc_manager.cpp
```diff
@@ -48,57 +55,80 @@ std::string MCCManager::NextHost()
...
+
+        DO_ASSERT(it != _mccHostsByOriginalHost.end());
     }
-    DoLogVerbose("Returning MCC host: [%s]", mccHostName.data());
+    DoLogVerbose("Returning MCC host: [%s] for original host: [%s]", mccHostName.data(), originalHost.data());
     return mccHostName;
 }
 
-bool MCCManager::NoFallback() const
+// Returns true if MCC host is banned, false otherwise
```
jimson-msft, line 84: Comment applies to IsBanned? and not ReportHostError correct? Consider deleting anyways cause the function signature explains it well enough

client-lite/src/config/mcc_manager.h
shishirb-MSFT: Yes, will change to class.

client-lite/src/config/mcc_manager.cpp
shishirb-MSFT: Yes, will delete. The method changed mid way through the changes :-)

client-lite/src/download/download.cpp
```diff
@@ -432,20 +444,135 @@ UINT Download::_MaxNoProgressIntervals() const
...
+std::string Download::_UpdateConnectionTypeAndGetUrl(bool retryAfterFailure)
+{
+    auto newConnectionType = _connectionType;
+    if (retryAfterFailure)
+    {
+        DO_ASSERT(_connectionType != ConnectionType::None);
+
+        _fAllowMcc = !_FallbackFromMccDue();
+
+        if (_fAllowMcc && !_mccFallbackDue)
```
jimson-msft, line 464: Can this be rewritten as:

if (!_FallbackfromMccDue())

Since _mccFallbackDue is already checked in 462

client-lite/src/download/download.cpp
```diff
@@ -432,20 +444,135 @@ UINT Download::_MaxNoProgressIntervals() const
...
+    {
+        _numAttemptsWithCurrentConnectionType++;
+    }
+
+    DoLogInfo("%s, connection type: %d, numAttempts: %u, url: %s",
+        GuidToString(_id).data(), static_cast<int>(_connectionType), _numAttemptsWithCurrentConnectionType, urlToUse.data());
+    return urlToUse;
+}
+
+bool Download::_ShouldFallbackFromMccNow(bool isFatalError) const
```
jimson-msft, line 532: >_ShouldFallbackFromMccNow [](http://example.com/codeflow?start=15&length=25)

Name is kind of confusing, consider _ShouldStopUsingMCC

client-lite/src/download/download.cpp
```diff
@@ -432,20 +444,135 @@ UINT Download::_MaxNoProgressIntervals() const
...
-        if (!_mccHost.empty() && _mccManager.NoFallback() && HttpAgent::IsClientError(_httpStatusCode))
+        return g_progressTrackerMaxNoProgressIntervals;
+    }
+
+    auto maxNoProgressIntervals = std::round(std::chrono::duration<double>(_noProgressTimeout) / g_progressTrackerCheckInterval);
+    DO_ASSERT(maxNoProgressIntervals >= 1);
+    return static_cast<UINT>(maxNoProgressIntervals);
+}
+
+std::string Download::_UpdateConnectionTypeAndGetUrl(bool retryAfterFailure)
```
jimson-msft, line 455: >_UpdateConnectionTypeAndGetUr [](http://example.com/codeflow?start=22&length=29)

So the logic here is to oscillate between cdn and mcc? The download will never time out?

jimson-msft (approved):
:shipit:

client-lite/src/download/download.cpp
shishirb-MSFT: Oscillate, yes. The no progress timeout is still enabled, no changes to _SchedProgressTracking().

client-lite/src/download/download.cpp
shishirb-MSFT: It's not a permanent switch to CDN so I'll rename to _ShouldPauseMccUsage().

client-lite/src/download/download.cpp
shishirb-MSFT: Reworked this a bit.

jimson-msft (approved):
:shipit:

PR #49, opened 2021-07-13
Use 1ES Self-Hosted Images provisioned with dependencies

jimson-msft:


shishirb-MSFT:
> # Have to use cmdline rather than built in python task, because the python env variable chooses python2

Is this true even with the 1ES images?

---
Refers to: azure-pipelines/build/templates/dopapt-native-steps.yml:28 in 3d35aa5. [](commit_id = 3d35aa528ec0448f0c8d7095ad79ebd5bf602f33, deletion_comment = False)

shishirb-MSFT (approved):
:shipit:

jimson-msft:
> # Have to use cmdline rather than built in python task, because the python env variable chooses python2

Unfortunately yes =-/, I had a commit where I tried out the python task and it auto selected python 2. There's another azdo pipeline task for specifying a target python version within a job, but that requires python installation to the tools directory of the self-hosted agent, which would require us to have an additional installation step.



---
In reply to: [879487839](https://github.com/microsoft/do-client/pull/49#issuecomment-879487839) [](http://example.com/codeflow?ancestors=879487839,879487839)

---
Refers to: azure-pipelines/build/templates/dopapt-native-steps.yml:28 in 3d35aa5. [](commit_id = 3d35aa528ec0448f0c8d7095ad79ebd5bf602f33, deletion_comment = False)

PR #51, opened 2021-07-14
Migrate azdo resources to 1ES managed pool

jimson-msft:


azure-pipelines/build/templates/doclient-lite-docker-steps.yml
```diff
@@ -2,18 +2,32 @@
 # Consume this steps template in one or more jobs by passing in parameter values.
 
 parameters:
-- name: targetOsArch    # example: debian10-arm32
+- name: targetOS        # example: ubuntu1804
```
shishirb-MSFT, line 5: >targetOS [](http://example.com/codeflow?start=8&length=8)

targetOS seems to be unused, no?

shishirb-MSFT: In all the docker steps templates.

azure-pipelines/publishing/github-release.yml
```diff
@@ -6,7 +6,12 @@ trigger:
 pr:
 - none
 
-pool: dotestlab
+#pool: 1es_hosted_pool_ubuntu_1804
```
shishirb-MSFT, line 9: I think we are ready to switch this one also to the 1ES pool right?

build/bootstrap.sh
```diff
@@ -0,0 +1,220 @@
...
+    # Installs to a non-standard location so add to PATH manually
+    export PATH=$PATH:~/.local/bin
+}
+
+# TODO(jimson): Running docker builds on 1ES hosted agents are currently blocked due to az pipeline agent not having permissions to call the docker daemon
+# Tried potential solutions here: https://docs.docker.com/engine/security/rootless/, to no avail.
+# 1. Using 'usermod -aG docker $USER' hangs the az pipeline agent on 'newgrp docker', which is required for the permission changes to take effect
+# 2. The rootless install succeeds, but azdo pipeline agent still reports permissions failure when calling into docker daemon
+# See if we can leverage someone elses docker install task from Image Factory when provisioning the 1ES managed image
+# After docker is callable on a 1ES managed image, we can swap all pipelines to use 1ES hosted pool instead of the microsoft hosted agent, and we can remove usage of the bootstrap script
```
jimson-msft, line 159: This comment can now be deleted

azure-pipelines/build/templates/doclient-lite-docker-steps.yml
jimson-msft: Thanks good catch - after removing the install script steps, targetOS is now unused. Need to remove from here as well as the non-template yaml for docker as well.

azure-pipelines/publishing/github-release.yml
jimson-msft: Yup!

build/bootstrap.sh
shishirb-MSFT: The lines above, about permissions issue calling into docker daemon, can be removed also, right?

How did you resolve it?

jimson-msft: Yes, I added comment on codeflow - but I think on github it only highlights the bottom-most line. The whole comment block can be removed.

So this comment came out of my first PR for this work, where I was using an Ubuntu1804 base image with no dependencies installed. The issue was that when trying to install docker via pipeline step - the current user is running as the self-hosted agent, with network service permissions. As a result - the pipeline agent would fail to perform certain commands, like pulling from a remote container registry.
When running this script as an ImageFactory artifact - docker gets installed with the right permissions to have the docker CLI callable through the pipeline agent's user. 

So the tl;dr is that it self-resolved once I did the install as a part of image provisioning rather than trying to install docker as a pipeline step.

build/bootstrap.sh
```diff
@@ -0,0 +1,220 @@
...
+
+# Defaults 
+INSTALL=all
+
+function usage {
+    cat <<EOM
+$(basename $0) - Script to setup development environments for Delivery Optimization
+Usage: $(basename $0) --platform <platform to install for> --install <install command>
+    --platform          # Platform to provision, supported platforms: ubuntu1804, ubuntu2004, debian9, debian 10. Required
+    --install           # Which command to run, supported commands: builddependencies, developertools, containertools, qemu, all. Default is all
```
jimson-msft, line 18: >dependencies [](http://example.com/codeflow?start=73&length=12)

fix this to just read 'build'

azure-pipelines/build/templates/doclient-lite-docker-steps.yml
```diff
@@ -2,18 +2,30 @@
...
 - name: config          # debug/release/minsizerel/relwithdebuginfo
   type: string
 
 steps:
+- task: Docker@2
+  displayName: Login to ACR
+  inputs:
+    command: login
+    containerRegistry: doclientcontainerregistry
+    repository: $(parameters.targetOsArch)
```
shishirb-MSFT, line 16: >$(parameters.targetOsArch) [](http://example.com/codeflow?start=16&length=26)

You will need to rename the repositories on azure portal now to change from dash to underscore.

shishirb-MSFT: The github release yaml is still using dash (ubuntu1804-x64).

IIRC, pipeline job names do not accept dash which is the reason we had the OS and arch as separate and used dash/underscore as appropriate.

azure-pipelines/build/doclient-lite-docker.yml
```diff
@@ -14,26 +14,21 @@ schedules:
...
 
-pool: dotestlab
+pool: 1es_hosted_pool_ubuntu_1804
 
 jobs:
 - template: templates/do-docker-jobs.yml
   parameters:
-    targetOS: 'debian10'
+    targetOsArch: 'debian9_arm32'
     targetArch: 'arm32'
```
shishirb-MSFT, line 23: >targetArch: 'arm32' [](http://example.com/codeflow?start=4&length=19)

If you don't go back to separating the OS and arch, remove targetArch here.

azure-pipelines/build/templates/doclient-lite-docker-steps.yml
jimson-msft: Correct, saw the dash issue in pipeline job names as well. Let's use the underscore convention everywhere. 

Re: repositories on azure portal - do you mean the container repository?

azure-pipelines/build/doclient-lite-docker.yml
jimson-msft: missed this one

azure-pipelines/build/templates/doclient-lite-docker-steps.yml
jimson-msft: renamed to all use underscore, and just use a single parameter targetOsArch:
[https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/22a3997b-5698-4d18-b7f5-0d2b6af877bf/resourceGroups/do-client/providers/Microsoft.ContainerRegistry/registries/DOClientContainerRegistry/repository](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/22a3997b-5698-4d18-b7f5-0d2b6af877bf/resourceGroups/do-client/providers/Microsoft.ContainerRegistry/registries/DOClientContainerRegistry/repository)

shishirb-MSFT:
>         targetOsArch: 'ubuntu1804-x64'

dash to underscore in all jobs here too.

---
Refers to: azure-pipelines/publishing/github-release.yml:31 in fb134bb. [](commit_id = fb134bb977ac2e77acd909df9fe99e9428a06cba, deletion_comment = False)

shishirb-MSFT (dismissed):
:shipit:

shishirb-MSFT (approved):
:shipit:

PR #53, opened 2021-07-15
Use ninja for arm builds

jimson-msft:


azure-pipelines/build/templates/doclient-lite-docker-steps.yml
```diff
@@ -24,7 +24,6 @@ steps:
 
 - task: CmdLine@2
   inputs:
-  # Unix Makefiles used in place of Ninja due to issues with 32-bit compatability on cross-arch builds
```
shishirb-MSFT, line 27: >Makefiles [](http://example.com/codeflow?start=9&length=9)

So the agent was already using ninja?

jimson-msft: Apparently, yes

shishirb-MSFT (approved):
:shipit:

PR #54, opened 2021-08-20
Agent: Replace boost.log usage with custom logger

shishirb-MSFT:


client-lite/src/ipc/rest_api_request.cpp
```diff
@@ -134,7 +134,6 @@ HRESULT RestApiDownloadStateChangeRequest::ParseAndProcess(DownloadManager& down
 HRESULT RestApiGetStatusRequest::ParseAndProcess(DownloadManager& downloadManager, RestApiParser& parser,
     web::json::value& responseBody)
 {
-    DoLogInfo("");
```
jimson-msft, line 137: nice catch, this was doing nothing before?

client-lite/src/trace/do_log.h
```diff
@@ -1,24 +1,36 @@
...
+namespace DOLog
+{
 
-void LogRuntimeFailure(const docli::FailureInfo& failure) noexcept;
+enum class Level
+{
+    Error,
+    Warning,
+    Info,
+    Verbose,
```
jimson-msft, line 13: >, [](http://example.com/codeflow?start=11&length=1)

nit: does last one need comma?

client-lite/src/trace/do_log.h
```diff
@@ -1,24 +1,36 @@
...
+#define EVENT_LEVEL_VERBOSE     DOLog::Level::Verbose
 
 // '##' is required before __VA_ARGS__ to allow an empty arg list to the variadic macro.
 // MSVC supports this by default but GCC requires '##'. C++2a has added VA_OPT macro
 // to officially support this behavior.
 
-#define DoLogMessage(level, msg, ...)   ::LogMessage((level), __FUNCTION__, __LINE__, (msg), ##__VA_ARGS__)
-#define DoLogResult(lvl, hr, msg, ...)  ::LogResult((lvl), __FUNCTION__, __LINE__, (hr), (msg), ##__VA_ARGS__)
+#define DoLogMessage(level, msg, ...)   DOLog::Write((level), __FUNCTION__, __LINE__, (msg), ##__VA_ARGS__)
+#define DoLogResult(lvl, hr, msg, ...)  DOLog::WriteResult((lvl), __FUNCTION__, __LINE__, (hr), (msg), ##__VA_ARGS__)
```
jimson-msft, line 33: >lvl [](http://example.com/codeflow?start=20&length=3)

nit: should this be level, for consistency with line 32?

client-lite/src/trace/do_log.cpp
```diff
@@ -0,0 +1,437 @@
...
+        const size_t cchBuffer = BytesFree();
+
+        // Timestamp ProcessID ThreadID severity
+        HRESULT hr = StringPrintf(pszBuffer, cchBuffer, &cchWritten, "%s %-5d %-5d %-8s ", timeStr.data(), pid, tid, pszLevel);
+        if (SUCCEEDED(hr))
+        {
+            cchTotalWritten += cchWritten;
+
+            // Note: pszFunc may not be available. Example is logging from error_macros.cpp in release builds.
+            if (hrIn != HRESULT(-1))
```
jimson-msft, line 106: >-1 [](http://example.com/codeflow?start=32&length=2)

S_OK is 0 no?

client-lite/src/trace/do_log.cpp
```diff
@@ -0,0 +1,437 @@
...
+            std::unique_lock<std::mutex> lock(_mutex);
+            _cv.wait_for(lock, std::chrono::minutes(3), [this](){ return !_fRunning; });
+            if (!_fRunning)
+            {
+                break;
+            }
+
+            _writeBuffer.Flush(_logFile);
+            _RotateLogFilesIfNeeded();
+            ++_stats.numFlushThreadLoops;
```
jimson-msft, line 291: Would it also be beneficial to have a max memory limit for logging (incase of agent business -> logging a ton of stuff?) i.e. flush if we're using x memory

jimson-msft:
> 

eventually will this make it's way to the common/ folder to be used for the plugin logging like you had mentioned?

---
Refers to: client-lite/src/trace/do_log.h:50 in 5f4a276. [](commit_id = 5f4a276bd8e06023dd56518ccefccee71dad45fe, deletion_comment = False)

client-lite/src/trace/do_log.cpp
```diff
@@ -0,0 +1,437 @@
...
+            failure.uLineNumber);
+    }
+    else
+    {
+        WriteResult(Level::Error, failure.pszFunction, failure.uLineNumber, failure.hr, "[%s, %d]", failure.pszFile, failure.uLineNumber);
+    }
+}
+
+// LoggerImpl destructor performs non-trivial tasks like waiting for thread termination.
+// Use raw pointer to avoid executing the destructor on process termination if Close() was not invoked.
```
jimson-msft, line 394: This is to avoid hanging the process for the potential 3 min wait time correct? Don't we lose the last bit of logging if the process terminates early (i.e. manually for example with systemctl restart on the agent daemon?)

jimson-msft (approved):
:shipit:

client-lite/src/ipc/rest_api_request.cpp
shishirb-MSFT: I think the intention was to log the entry to this method, which it does. However, gcc only provides the method name and not the class name so it is not so useful here.

client-lite/src/trace/do_log.h
shishirb-MSFT: No but it's a convention to allow cleaner diff when adding a new entry - no need to modify existing lines of code.

shishirb-MSFT:
> 

Yes that's what I have in mind but I'm also wondering whether the plugin needs such elaborate logging. The plugin also has the complication of potentially running multiple instances concurrently (apt can invoke multiple instances of the plugin concurrently).

---
In reply to: [902842051](https://github.com/microsoft/do-client/pull/54#issuecomment-902842051) [](http://example.com/codeflow?ancestors=902842051,902842051)

---
Refers to: client-lite/src/trace/trace_src.h:50 in 5f4a276. [](commit_id = 5f4a276bd8e06023dd56518ccefccee71dad45fe, deletion_comment = False)

client-lite/src/trace/do_log.cpp
shishirb-MSFT: -1 here indicates there is no HRESULT to log. See the calls to this method in Write() and WriteResult() below.

client-lite/src/trace/do_log.cpp
shishirb-MSFT: Yes this is done already. The LogBuffer is a fixed size buffer (64KB right now). LoggerImpl::WriteV method performs the flush in case buffer gets full within this 3 min interval.

client-lite/src/trace/do_log.cpp
shishirb-MSFT: You're right on the hanging part, although I don't know if that applies to linux (windows will definitely hang if we wait for a thread to terminate within the process termination flow). I'll see if I can simplify this later.

All normal process termination will invoke Close() and thus the LoggerImpl destructor will peform the final flush. So we'll lose data only if the process crashes or is killed externally.

client-lite/src/trace/do_log.h
shishirb-MSFT: Yes, will do.

jimson-msft:
> 

True, maybe not needed then. If the plugin can spin up multiple instances then the cleanup logic may be problematic if only allowing 3 log files

---
In reply to: [902850523](https://github.com/microsoft/do-client/pull/54#issuecomment-902850523) [](http://example.com/codeflow?ancestors=902850523,902842051)

---
Refers to: client-lite/src/trace/trace_src.h:50 in 5f4a276. [](commit_id = 5f4a276bd8e06023dd56518ccefccee71dad45fe, deletion_comment = False)

shishirb-MSFT:
> 

Exactly. I'm going to discuss with JeffM if he has any suggestions for lightweight logging on linux (syslog?).

---
In reply to: [902857112](https://github.com/microsoft/do-client/pull/54#issuecomment-902857112) [](http://example.com/codeflow?ancestors=902857112,902842051)

---
Refers to: client-lite/src/trace/trace_src.h:50 in 5f4a276. [](commit_id = 5f4a276bd8e06023dd56518ccefccee71dad45fe, deletion_comment = False)

PR #64, opened 2021-08-31
Pre-cpprest-removal: Remove dependency in REST API parsing

shishirb-MSFT:


common/lib-dohttp/do_cpprest_base_uri.h
```diff
@@ -0,0 +1,394 @@
...
+{
+namespace deliveryoptimization
+{
+namespace details
+{
+namespace cpprest_web
+{
+
+/*
+    Code borrowed from cpprestsdk project: https://github.com/microsoft/cpprestsdk/
```
shishirb-MSFT, line 20: >Code borrowed [](http://example.com/codeflow?start=4&length=13)

@jimson-msft I've copied the code as is for the most part. Only differences are in the types (no support for wide strings) and namespaces.

common/lib-dohttp/do_cpprest_base_uri.h
```diff
@@ -0,0 +1,394 @@
...
+    libcurl includes APIs to work with URLs starting from v7.62.0.
+    Ubuntu 18.04 ships with libcurl v7.58.0 only so we cannot use the URL API.
```
jimson-msft, line 22: That's unfortunate :-(

Unrelated to this pr but I recently read somewhere that the fragmentation of the linux world - specifically with library and version support is what made windows so dominant for game developers - unified apis for graphics/visuals/audio on a single system with great backcompat. Very few people wanted to develop libraries to support linux distros, esp with PC market share being relatively low

We've been feeling the pain of it as well with dependency management and platform support for all these diff distros.

shishirb-MSFT: Ah, I see. Yeah it is really unfortunate.

PR #65, opened 2021-08-31
Remove cpprest dependency in REST listener

shishirb-MSFT:


client-lite/src/ipc/rest_http_listener_conn.cpp
```diff
@@ -0,0 +1,139 @@
...
+#define RETURN_HTTP_STATUSCODE_STR(code) \
+    case msdod::http_status_codes::code: return #code;
+
+    switch (statusCode)
+    {
+        RETURN_HTTP_STATUSCODE_STR(OK);
+        RETURN_HTTP_STATUSCODE_STR(BadRequest);
+        RETURN_HTTP_STATUSCODE_STR(NotFound);
+        RETURN_HTTP_STATUSCODE_STR(InternalError);
+        RETURN_HTTP_STATUSCODE_STR(ServiceUnavailable);
```
jimson-msft, line 60: If it's not one of these 5, then the http status code is set to "StatusDescription" in HttplistenerConnection::Reply(), is this intended?

client-lite/src/ipc/rest_http_listener_conn.cpp
```diff
@@ -0,0 +1,139 @@
...
+void HttpListenerConnection::Reply(unsigned int statusCode, const std::string& body)
+{
+    std::stringstream ss;
+    ss << "HTTP/1.1 " << statusCode << ' ' << g_HttpStatusToString(statusCode) << "\r\n";
+    if (!body.empty())
+    {
+        ss << "Content-Length: " << body.size() << "\r\n";
+    }
+    ss << "Server: Delivery-Optimization-Agent/" << microsoft::deliveryoptimization::util::details::SimpleVersion() << "\r\n";
+    ss << "\r\n";
```
jimson-msft, line 75: This crlf is used for the empty header correct?

client-lite/src/ipc/rest_http_listener_conn.cpp
shishirb-MSFT: Yes, we currently don't use anything else. The description actually doesn't matter in the protocol since it is purely for humans to read.

client-lite/src/ipc/rest_http_listener_conn.cpp
shishirb-MSFT: Yes, it signifies the end of the header section.

client-lite/src/ipc/rest_http_listener.cpp
```diff
@@ -36,25 +34,52 @@ void RestHttpListener::Start(const std::string& listenUrl)
...
+
+void RestHttpListener::_BeginAccept()
+{
+    auto acceptSocket = std::make_shared<boost_tcp_t::socket>(*_io);
+    _listener->async_accept(*acceptSocket, [this, acceptSocket](const boost::system::error_code& ec) mutable
+        {
+            if (ec)
+            {
+                DoLogError("Async accept failed, error = %d", ec.value());
+                // TODO(shishirb) terminate process by throwing exception?
```
jimson-msft, line 72: Would it make sense to try again up until n failures?

jimson-msft (approved):
:shipit:

client-lite/src/ipc/rest_http_listener.cpp
shishirb-MSFT: Maybe. The only time I've seen a failure here is during shutdown notifying us that socket is closed (expected). So I'll leave it as is until we see other retry-able errors.

PR #66, opened 2021-09-02
Replace entire cpprest usage with libcurl

shishirb-MSFT:


NOTICE.md
```diff
@@ -17,46 +17,6 @@ required to debug changes to any libraries licensed under the GNU Lesser General
 
 ====================================================================
 
-microsoft/cpprestsdk 18212a2a7967e12d740bfb957e500892b3463c88 - MIT
```
shishirb-MSFT, line 20: >cpprestsdk [](http://example.com/codeflow?start=10&length=10)

I'll follow up with Carmen about adding libcurl info here and in cgmanifest.

client-lite/src/download/download.cpp
```diff
@@ -1,6 +1,7 @@
 #include "do_common.h"
 #include "download.h"
 
+#include <cmath>
```
jimson-msft, line 4: is this unused?

shishirb-MSFT: It is used for std::round on line 455. I think it was being brought in by one of cpprest's headers earlier.

build/bootstrap.sh
```diff
@@ -93,29 +83,9 @@ function installBuildDependencies
     fi
 
     # Open-source library dependencies
-    # Boost libs for DO
+    # Boost libs
```
jimson-msft, line 86: Can just delete this comment now. Before it was to distinguish which were used for DO and which were used for cpprest

client-lite/src/util/http_agent.cpp
```diff
@@ -59,40 +62,61 @@ bool HttpAgent::ValidateUrl(const std::string& url)
...
+        _requestContext.requestHeaders = tempList;
+        curl_easy_setopt(_requestContext.curlHandle, CURLOPT_HTTPHEADER, _requestContext.requestHeaders);
+    }
+
+    _requestContext.responseHeaders.clear();
+    _requestContext.responseStatusCode = 0;
+    _requestContext.hrTranslatedStatusCode = S_OK;
+    _requestContext.responseOnHeadersAvailableInvoked = false;
+    _requestContext.responseOnCompleteInvoked = false;
+    _callbackContext = callerContext;
```
jimson-msft, line 89: >callerContext; [](http://example.com/codeflow?start=23&length=14)

Do we need to expose this as a function param? Looks like it's never set and default value is always used. It looks like a param for the IHttpEvents interface but unused by the functions implementing that interface

client-lite/src/util/http_agent.cpp
```diff
@@ -157,144 +184,50 @@ HRESULT HttpAgent::_CreateClient(PCSTR szUrl, PCSTR szProxyUrl) try
...
-
-        web::http::uri remoteUri(url);
-        _client = std::make_unique<web::http::client::http_client>(remoteUri, clientConfig);
-        DoLogVerbose("New http_client for %s", remoteUri.to_string().data());
+        _requestContext.curlHandle = curl_easy_init();
+        RETURN_HR_IF(E_OUTOFMEMORY, _requestContext.curlHandle == nullptr);
+        DoLogVerbose("New http_client for %s", szUrl);
+
+        // Set options that are generic to all requests
+        curl_easy_setopt(_requestContext.curlHandle, CURLOPT_FOLLOWLOCATION, static_cast<long>(1));
```
jimson-msft, line 199: >curl_easy_setopt(_requestContext.curlHandle, CURLOPT_FOLLOWLOCATION, static_cast<long>(1)); [](http://example.com/codeflow?start=8&length=91)

Maybe I misunderstood but you mentioned in sync 302'ing urls needed manual handling? Doesn't this take care of redirects?

client-lite/src/util/http_agent.cpp
```diff
@@ -157,144 +184,50 @@ HRESULT HttpAgent::_CreateClient(PCSTR szUrl, PCSTR szProxyUrl) try
...
+        curl_easy_setopt(_requestContext.curlHandle, CURLOPT_MAXREDIRS, static_cast<long>(10));
```
jimson-msft, line 200: >10 [](http://example.com/codeflow?start=90&length=2)

-1 allows infinite redericts, 10 to avoid infinite redirecting (can such a thing happen?)

jimson-msft (approved):
:shipit:

client-lite/src/exe/docs.cpp
```diff
@@ -118,14 +120,31 @@ class BoostAsioService
...
 HRESULT Run() try
 {
     InitializeDOPaths();
 
+    CurlGlobalInitCall curlGlobalInit;
     BoostAsioService asioService;
 
     ConfigManager clientConfigs;
-    auto downloadManager = std::make_shared<DownloadManager>(clientConfigs);
+    CurlMultiOperation curlOperations;
```
shishirb-MSFT, line 146: >curlOperations [](http://example.com/codeflow?start=23&length=14)

Move this to DownloadManager.

client-lite/src/util/http_agent.cpp
shishirb-MSFT: This will make libcurl follow the redirect automatically but we still get headers back for the 302 response before getting headers for the 200.

client-lite/src/util/http_agent.cpp
shishirb-MSFT: Will remove in the next PR.

client-lite/src/util/do_curl_multi_operation.cpp
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    if (wrappedHandleToBeDeleted)
+    {
+        // outside the lock, wait for the handle to be removed from _activeHandles
+        (void)wrappedHandleToBeDeleted->inactiveSignal.Wait(std::chrono::minutes(30));
+    }
+}
+
+void CurlMultiOperation::_WorkerThread()
```
jimson-msft, line 75: Consider renaming this to _Work(),

client-lite/src/util/do_curl_multi_operation.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+        for (const auto& h : _handlesToRemove)
+        {
+            _activeHandles.Remove(h, _multiHandle);
+        }
+        _handlesToRemove.clear();
+
+        if (_activeHandles.Empty())
+        {
+            _cv.wait(lock, [this](){ return !_fKeepRunning || !_handlesToAdd.empty(); });
```
jimson-msft, line 99: >!_handlesToAdd.empty() [](http://example.com/codeflow?start=62&length=22)

if this is non-empty, wouldn't cv wait here forever? There's no way to drain/clear the vector except for line 89 no?

client-lite/src/util/do_curl_multi_operation.cpp
```diff
@@ -0,0 +1,247 @@
...
+            selectResult = select(maxfd + 1, &fdsRead, &fdsWrite, &fdsException, &timeout);
+        }
+
+        switch (selectResult)
+        {
+            case -1:
+                DO_ASSERT(false); // TODO(shishirb) how to handle?
+                break;
+
+            case 0:
```
jimson-msft, line 153: I think you might have mentioned this when we went over this, but for case 0: it indicates the process timed out? Should a comment be added on why we leave this case empty?

shishirb-MSFT: 0 = timeout, yes. Note this case isn't empty but falls through to call curl_multi_perform. I've added some comments in the next iteration.

client-lite/src/util/do_curl_multi_operation.cpp
shishirb-MSFT: It is the opposite - wait() will return if _handlesToAdd becomes non-empty or _fKeepRunning becomes false.

Also note that AddHandle will wait if we were executing line 89-99 at that time (same mutex).

PR #67, opened 2021-09-07
Agent: Cleanup after consuming libcurl

shishirb-MSFT:


client-lite/src/exe/docs.cpp
```diff
@@ -119,26 +120,11 @@ class BoostAsioService
...
-        curl_global_cleanup();
-    }
-};
-
 HRESULT Run() try
 {
     InitializeDOPaths();
 
-    CurlGlobalInitCall curlGlobalInit;
+    CurlGlobalInit curlGlobalInit;
```
jimson-msft, line 127: Can't CurlGlobalInit be initialized in the DownloadManager constructor?

jimson-msft (approved):
:shipit:

client-lite/src/exe/docs.cpp
shishirb-MSFT: Since it is a global setup, it is safer to do it here rather than in DownloadManager where it would have to be done before CurlRequests is constructed. Easy to get it wrong there.

Issue #58, opened 2021-08-25
build/boostrap.sh lacks execute permissions and has several errors on Ubuntu 20.04


oottka:
When cloning and building the project on a fresh Ubuntu 20.04 VM, the [build/boostrap.sh](https://github.com/microsoft/do-client/blob/main/build/bootstrap.sh) script has several issues. Steps to reproduce:

- Create a new Ubuntu 20.04 VM
- Clone the project
- Run `./build/boostrap.sh --platform ubuntu2004`

The script has the following issues:

- [The documentation](https://github.com/microsoft/do-client#development-machine-setup) tells to run `cd build/boostrap`. This is obviously wrong.
- The script does not have execute permissions (need to do chmod u+x before it can be run)
- After setting execute permissions and running the script, the following error is thrown:
```
-- websocketpp not found, using the embedded version
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/Boost-1.71.0/BoostConfig.cmake:117 (find_package):
  Could not find a package configuration file provided by "boost_thread"
  (requested version 1.71.0) with any of the following names:

    boost_threadConfig.cmake
    boost_thread-config.cmake

  Add the installation prefix of "boost_thread" to CMAKE_PREFIX_PATH or set
  "boost_thread_DIR" to a directory containing one of the above files.  If
  "boost_thread" provides a separate development package or SDK, be sure it
  has been installed.
Call Stack (most recent call first):
  /usr/lib/x86_64-linux-gnu/cmake/Boost-1.71.0/BoostConfig.cmake:182 (boost_find_component)
  /usr/share/cmake-3.16/Modules/FindBoost.cmake:443 (find_package)
  Release/cmake/cpprest_find_boost.cmake:49 (find_package)
  Release/cmake/cpprest_find_websocketpp.cmake:17 (cpprest_find_boost)
  Release/src/CMakeLists.txt:68 (cpprest_find_websocketpp)


-- Configuring incomplete, errors occurred!
```
This can be fixed by installing libboost-thread-dev.
- After installing libboost-thread-dev and re-running the script, the following error is thrown:
```
Package python-pip is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python3-pip

E: Package 'python-pip' has no installation candidate
```
This is because python-pip no longer exists in Ubuntu 20.04, it being replaced by python3-pip. This can be fixed by modifying the script to install python3-pip on Ubuntu instead.

[...]

PR #68, opened 2021-09-08
Enable versioning for build images

jimson-msft:
- Docker build containers now support specifying version. This version aligns with the next github release version in semantic versioning format.
- Also specify vmImage name as the host machine needs to be patched to use libcurl instead of cpprest.

### DIFF
--- a/azure-pipelines/build/doclient-lite-docker.yml
+++ b/azure-pipelines/build/doclient-lite-docker.yml
@@ -20,7 +20,7 @@ variables:
 
 pool: 
   name: 1es_hosted_pool_ubuntu_1804
-  vmImage: do-adu-build-${{variables.version}}
+  vmImage: do-adu-build-$(variables.version)
 
 jobs:
 - template: templates/do-docker-jobs.yml
--- a/azure-pipelines/build/doclient-lite-docker.yml
+++ b/azure-pipelines/build/doclient-lite-docker.yml
@@ -27,18 +27,18 @@ pool:
 jobs:
 - template: templates/do-docker-jobs.yml
   parameters:
-    targetOsArch: 'debian9-arm32'
+    targetOsArch: 'debian9_arm32'
     version: ${{variables.version}}
     stepsTemplate: 'doclient-lite-docker-steps.yml'
 
 - template: templates/do-docker-jobs.yml
   parameters:
-    targetOsArch: 'debian10-arm32'
+    targetOsArch: 'debian10_arm32'
     version: ${{variables.version}}
     stepsTemplate: 'doclient-lite-docker-steps.yml'
 
 - template: templates/do-docker-jobs.yml
   parameters:
-    targetOsArch: 'ubuntu1804-arm64' # azure pipelines does not support '.' in display names
+    targetOsArch: 'ubuntu1804_arm64' # azure pipelines does not support '.' in display names
     version: ${{variables.version}}
     stepsTemplate: 'doclient-lite-docker-steps.yml'
--- a/azure-pipelines/build/doclient-lite-docker.yml
+++ b/azure-pipelines/build/doclient-lite-docker.yml
@@ -20,7 +20,9 @@ variables:
 
 pool: 
   name: 1es_hosted_pool_ubuntu_1804
-  vmImage: do-adu-build-$(version)
+  #TODO(jimson): After Imagefactory check-in is completed and we can create new image, use the commented below
+  #vmImage: do-adu-build-$(version)
+  vmImage: do-adu-build-2
 
 jobs:
 - template: templates/do-docker-jobs.yml
--- a/azure-pipelines/build/doclient-lite-docker.yml
+++ b/azure-pipelines/build/doclient-lite-docker.yml
@@ -1,6 +1,10 @@
 # Pipeline to build DO Agent using docker to target non-native OS and/or architecture.
 # Publishes the binaries + packages as artifacts.
 
+variables:
+- name: version
+  value: 0.8.0
+
 # Disable branch and pr triggers - currently run this manually to avoid hogging build machine resources
 # Rely on Ubuntu x64 pipeline for CI/CD
 trigger:
@@ -13,10 +17,6 @@ schedules:
   branches:
     include:
     - main
-
-variables:
-- name: version
-  value: 0.8.0
 
 pool: 
   name: 1es_hosted_pool_ubuntu_1804
--- a/azure-pipelines/build/doclient-lite-docker.yml
+++ b/azure-pipelines/build/doclient-lite-docker.yml
@@ -25,18 +25,18 @@ pool:
 jobs:
 - template: templates/do-docker-jobs.yml
   parameters:
-    targetOsArch: 'debian9_arm32'
+    targetOsArch: 'debian9-arm32'
     version: ${{variables.version}}
     stepsTemplate: 'doclient-lite-docker-steps.yml'
 
 - template: templates/do-docker-jobs.yml
   parameters:
-    targetOsArch: 'debian10_arm32'
+    targetOsArch: 'debian10-arm32'
     version: ${{variables.version}}
     stepsTemplate: 'doclient-lite-docker-steps.yml'
 
 - template: templates/do-docker-jobs.yml
   parameters:
-    targetOsArch: 'ubuntu1804_arm64' # azure pipelines does not support '.' in display names
+    targetOsArch: 'ubuntu1804-arm64' # azure pipelines does not support '.' in display names
     version: ${{variables.version}}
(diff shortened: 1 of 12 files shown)


jimson-msft:
@shishirb-MSFT Just a draft for your review for now, I'm going to wait until the other PR for imagefactory gets checked in so we can use do-adu-build-<version> rather than the old machine. As a result of still using the old machine, builds on this branch will fail (no libcurl).

Note: there are failures in deb9 builds resulting from std::bad_cast():
https://deliveryoptimization.visualstudio.com/client/_build/results?buildId=3256&view=logs&j=f64b1c2f-1b39-5c58-9002-e3b993974b94&t=520448b5-31ed-5301-4234-9625ec8fd8ff
https://deliveryoptimization.visualstudio.com/client/_build/results?buildId=3253&view=logs&j=f64b1c2f-1b39-5c58-9002-e3b993974b94&t=520448b5-31ed-5301-4234-9625ec8fd8ff
https://deliveryoptimization.visualstudio.com/client/_build/results?buildId=3249&view=logs&j=f64b1c2f-1b39-5c58-9002-e3b993974b94

azure-pipelines/build/doclient-lite-docker.yml
```diff
@@ -1,6 +1,10 @@
 # Pipeline to build DO Agent using docker to target non-native OS and/or architecture.
 # Publishes the binaries + packages as artifacts.
 
+variables:
+- name: version
```
shishirb-MSFT, line 5: >version [](http://example.com/codeflow?start=8&length=7)

Do we need this variable? Wouldn't we just update the version in a branch when it needs to be changed?

shishirb-MSFT:
How are the docker images updated with latest patches?

---
In reply to: [915645193](https://github.com/microsoft/do-client/pull/68#issuecomment-915645193) [](http://example.com/codeflow?ancestors=915645193,915645193)

shishirb-MSFT:
Also, I don't want dev local builds+testing to be different from pipeline build+testing as much as possible. So it would be nice if we can keep native builds.

---
In reply to: [915658658](https://github.com/microsoft/do-client/pull/68#issuecomment-915658658) [](http://example.com/codeflow?ancestors=915658658,915645193,915645193)

jimson-msft:
Docker images will still be manually patched if we need to make dependency changes.
There might be a way for us to setup pipelines which create new docker images we detect changes in bootstrap.sh

For native builds - there's no difference except the build itself happens inside of a docker image.
Everything else - such as other build output (test exes for sdk + client) are written to the host machine, and still ran from there. Packages are installed/removed on the host.

This solves the problem of having to update the artifact we own in Imagefactory everytime we make a change to bootstrap.sh (no need to update that image to install libcurl, because the build itself will run in a docker container which we can update easily).


---
In reply to: [915661193](https://github.com/microsoft/do-client/pull/68#issuecomment-915661193) [](http://example.com/codeflow?ancestors=915661193,915645193)

shishirb-MSFT:
re:patches - I meant security/OS patches. The reason I'm asking is the whole point of moving to 1ES was to let them take care of VM/image patching.

---
In reply to: [915665454](https://github.com/microsoft/do-client/pull/68#issuecomment-915665454) [](http://example.com/codeflow?ancestors=915665454,915645193)

jimson-msft:
I'm not sure that same concern applies for containers. The container image is secure because it's just that - an image. Whereas a vm running a hosted agent has various attack vectors for bad actors to take advantage of, and it's connected to a network. 
We garuntee that by using an image associated with our ACR instance that when that image is spun into a container, we know exactly what's on it.
The host machine that is running the containers is still automatically managed by 1ES and patched when needed automatically

---
In reply to: [915674475](https://github.com/microsoft/do-client/pull/68#issuecomment-915674475) [](http://example.com/codeflow?ancestors=915674475,915645193)

shishirb-MSFT:
S360 has recently flagged security issues in our MCC instances also which uses docker containers. This is when we had to ask MCC team to update the container from 1804 to 2004 ubuntu.
Maybe the tooling hasn't caught up to scanning docker images on 1ES builds but I'm sure we'll get there.

---
In reply to: [915679890](https://github.com/microsoft/do-client/pull/68#issuecomment-915679890) [](http://example.com/codeflow?ancestors=915679890,915645193)

shishirb-MSFT:
What about runtime dependencies when running tests on the host VM? We'll need libcurl installed at runtime as well.

---
In reply to: [916332369](https://github.com/microsoft/do-client/pull/68#issuecomment-916332369) [](http://example.com/codeflow?ancestors=916332369,915679890,915645193)

azure-pipelines/build/doclient-lite-docker.yml
jimson-msft: Looks like github comment isnt showing up in codeflow:

It's needed in order to specify which docker image version to use. I can't think of a way to extract version #58 
With this approach, the latest development branch will always have this version set to the next release version

shishirb-MSFT: I meant can't we hardcode it in the yaml file itself? When will we need to update the version on the fly just before scheduling a pipeline run?

jimson-msft:
I was thinking about that too - the test run here passes:
[https://deliveryoptimization.visualstudio.com/client/_build/results?buildId=3262&view=logs&j=e76d410b-906a-5a54-f5d2-89a839293c70&t=242881bb-7330-5492-b5ab-fe5d3f311053](https://deliveryoptimization.visualstudio.com/client/_build/results?buildId=3262&view=logs&j=e76d410b-906a-5a54-f5d2-89a839293c70&t=242881bb-7330-5492-b5ab-fe5d3f311053)
The reason it's able to run is because libcurl is brought in by task "Install agent Debian package". Same with other runtime dependencies.

For test binaries - we could link to all libraries statically (though I'm not 100% sure if package-installed libraries all have static counterparts installed alongside). This for the case where we may use some dependency in test code that isn't used by the component itself.

re: container security - MCC is an actively running container on an active device. With the pipeline runs we have an isolated environment where the image (garunteed to be good, since its hosted on our ACR - unless ACR gets compromised), is pulled, then used within the run.
After the run completes, the container is decomissioned. The key here I think is if the host vm can get compromised during the duration of the run, which 1ES will handle for us.

Still, it's a good point. I'll spend some time reaching out to the right people or reviewing the 1ES migration notes to see if there's anything more specific on this.

---
In reply to: [916335207](https://github.com/microsoft/do-client/pull/68#issuecomment-916335207) [](http://example.com/codeflow?ancestors=916335207,915645193)

azure-pipelines/build/doclient-lite-docker.yml
jimson-msft: For the case where we need to patch a previous release. For example if we needed to patch v0.7.0 which uses cpprest - we can now specify it through the pipeline to use this set of docker images to do our builds.

shishirb-MSFT: Okay consider renaming to something more specific. vmImageVersion?

jimson-msft: maybe dependencyVersion. it might be used for both the docker image version & for specifying the image of the vmImage within the pool

shishirb-MSFT: buildEnvVersion?

jimson-msft: i think imageVersion is fine - referring to both vm & docker images. Added a comment next to it as well.

Waiting on the PR completion from ImageFactory and seeing runs go through before checking this in.

shishirb-MSFT (approved):
:shipit:

jimson-msft:
Discussed offline:
Since we are not shipping a container image or using the container in a production environment, we will not get flagged by s360.
For our use case, the guidance is to update these container images as a best practice - the same way we would for any vm.

For now, I will manually patch container images.
Opening this task to create pipelines that periodically patch all our container images:
[https://microsoft.visualstudio.com/OS/_workitems/edit/35945036](https://microsoft.visualstudio.com/OS/_workitems/edit/35945036)

---
In reply to: [916339889](https://github.com/microsoft/do-client/pull/68#issuecomment-916339889) [](http://example.com/codeflow?ancestors=916339889,916335207,915645193)

azure-pipelines/build/doclient-lite-docker.yml
jimson-msft: PR is completed, there appears to be an issue with updating the image payload within cloudtest - as a result, the reported new do-adu-build-0.8.0 image does not have libcurl installed, despite imagefactory logs showing libcurl being installed during image provisioning.

I'm going to complete the PR for now, but we won't see green builds until we get a response from the imagefactory team.

PR #69, opened 2021-09-10
Add libcurl to NOTICE and cgmanifest

shishirb-MSFT:


shishirb-MSFT:
@cmknox @jimson-msft The text in the NOTICE file was copied from https://curl.se/docs/copyright.html.
Is this the approach you took for the other libs as well?

jimson-msft:
I believe so, Carmen to verify.

---
In reply to: [916693720](https://github.com/microsoft/do-client/pull/69#issuecomment-916693720) [](http://example.com/codeflow?ancestors=916693720,916693720)

jimson-msft (dismissed):
:shipit:

jimson-msft (approved):
:shipit:

PR #74, opened 2021-09-16
Build fixes post Windows/Mac SDK support

shishirb-MSFT:


CMakeLists.txt
```diff
@@ -20,22 +20,60 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # Get verbose output from cmake generation and build steps
 set(CMAKE_VERBOSE_MAKEFILE ON)
 
+# Set target OS platform
+if (CMAKE_SYSTEM_NAME MATCHES Linux)
+    set(DO_PLATFORM_LINUX 1)
```
shishirb-MSFT, line 25: >DO_PLATFORM_LINUX [](http://example.com/codeflow?start=8&length=17)

Thought I'd simplify this while fixing the build issue. Easier/more correct to check for a definition of var/macro instead of its value.

shishirb-MSFT:
Pipelines all pass (client ARM still running but local build was fine).

shishirb-MSFT:
Trying the windows build now.

---
In reply to: [920607195](https://github.com/microsoft/do-client/pull/74#issuecomment-920607195) [](http://example.com/codeflow?ancestors=920607195)

build/build.py
```diff
@@ -485,8 +485,8 @@ def generator(self):
     @property
     def generate_options(self):
         return super().generate_options + [
-                f'-DCMAKE_TOOLCHAIN_FILE={get_vcpkg_toolchain_file_path(self.vcpkg_root_path)}',
-                f'-DVCPKG_TARGET_TRIPLET={self._vcpkg_triplet}'
+                '-DCMAKE_TOOLCHAIN_FILE={}'.format(get_vcpkg_toolchain_file_path(self.vcpkg_root_path)),
+                '-DVCPKG_TARGET_TRIPLET={}'.format(self._vcpkg_triplet)
```
jimson-msft, line 489: Was there an issue with the f string?

jimson-msft: Can't remember but is it because a version of python within arm builds is too old to support fstrings?

jimson-msft: nit: fix commented lines on 498-502, or consider removing them

sdk-cpp/include/do_config.h
```diff
@@ -1,7 +1,10 @@
 #ifndef _DELIVERY_OPTIMIZATION_DO_CONFIG_H
 #define _DELIVERY_OPTIMIZATION_DO_CONFIG_H
 
-#if (DO_CLIENT_ID == DO_CLIENT_ID_AGENT)
+//TODO(jimson): Callers may not have defined these compile definitions, as a result their builds may fail if the definition is not set when using the SDK
+//Look into removing platform specific header files from the SDK installation
+#if defined(DO_CLIENT_AGENT)
```
jimson-msft, line 6: It's tricky because I don't know if we'd ever intend to call these apis for non-agent use cases.

I think this is a potential way forward:
- Modify the sdk cmakelists to only install the right headers (depending on platform & agent), so the .deb file doesn't try to install do_download_property.h for now.
Otherwise, we can implement do_download_property for non-com use cases (will need to do this soon for mac anyways)
- In addition to the above, leave it up to callers who are consuming as source to add these definitions to their builds.

sdk-cpp/tests/test_data.cpp
```diff
@@ -20,10 +20,10 @@ const std::chrono::seconds g_smallFileWaitTime = 10s;
...
-#if (DO_INTERFACE_ID == DO_INTERFACE_ID_REST)
-const std::string g_docsProcName = "docs";
-const std::string g_docsSvcName = "do-client-lite.service";
-#elif (DO_INTERFACE_ID == DO_INTERFACE_ID_COM)
+#if defined(DO_INTERFACE_REST)
+const std::string g_docsProcName = "deliveryoptimization-agent";
+const std::string g_docsSvcName = "deliveryoptimization-agent.service";
+#elif defined(DO_INTERFACE_COM)
 const std::string g_smallFilePhfInfoJson = "{\"PiecesHashFileUrl\":\"https://eus2intdo101.blob.core.windows.net/partnercatalogcontainer/pc00051365650774478716C3536617271346B6B33563946436D6C794B6A516B2F74376751506F7A4933337472593D0000B9CCB33-51FB-4062-AFAC-08C8A088CB35.json\",\"HashOfHashes\":\"Q6VPwDxql56arq4kk3V9FCmlyKjQk/t7gQPozI33trY=\"}";
 #endif
```
jimson-msft, line 28: Oops, must have messed up the merge conflict resolution here - resulting in the client test failures

jimson-msft (approved):
:shipit:

jimson-msft:
@shishirb-MSFT thanks for fixing!

build/build.py
shishirb-MSFT: Yeah deb9 python is too old to support it.

sdk-cpp/include/do_config.h
shishirb-MSFT: Agreed that a compile def that is protecting an  entire .h/.cpp file should be handled via the build system instead. I'll leave that you for later.

shishirb-MSFT: Also agree that do_download_property should be implemented for all platforms/clients.

PR #75, opened 2021-09-16
Replace cpprestsdk usage with libcurl

shishirb-MSFT:


client-lite/CMakeLists.txt
```diff
@@ -33,16 +33,8 @@ endfunction ()
...
-# 1. v2.10.10 min version (see required PR link below). cpprestsdk does not seem to support specifying this through cmake.
-# https://github.com/microsoft/cpprestsdk/pull/1019/files
-#
-# 2. Installing libcpprest-dev via apt installs cpprest's cmake config files to a non-default cmake search path before v2.10.9
-# See: https://github.com/microsoft/cpprestsdk/issues/686
-# This issue has been patched but has not made its way to the stable branch for Ubuntu
-# Since we are statically linking to v2.10.16 we no longer need to worry about the above as cpprest is patched to provide the proper package configuration metadata
-find_package(cpprestsdk CONFIG REQUIRED)
+find_package(CURL REQUIRED)
+find_package(Threads REQUIRED)
```
jimson-msft, line 37: nit: consider adding a comment here for which header this is required for

licenses/cgmanifest.json
```diff
@@ -66,6 +56,42 @@
...
+          "Name": "libcurl4",
+          "Version": "7.64.0-4+deb10u2",
+          "Distribution": "Debian",
+          "Release": "10.10"
+        }
+      },
+      "DevelopmentDependency": false
+    },
     {
       "Component": {
```
jimson-msft, line 96: just wondering - do we need to also include the threads library here?

jimson-msft (approved):
:shipit:

client-lite/CMakeLists.txt
shishirb-MSFT: Not any particular header but a requirement from g++ when we use any threads in the project.

licenses/cgmanifest.json
shishirb-MSFT: No idea. Is it considered a 3rd party lib? We don't list c/c++ libs here either.

PR #76, opened 2021-09-16
Plugin: Log to std::cerr. File log only in debug mode.

shishirb-MSFT:


jimson-msft:
Sounds good to me.

---
In reply to: [921139047](https://github.com/microsoft/do-client/pull/76#issuecomment-921139047) [](http://example.com/codeflow?ancestors=921139047,921139047)

jimson-msft:
re: disk space concern - these files are pretty minimal and logging only occurs from plugin when apt instructs a download - could another option be to log on both retail & debug builds, but to clear the oldest files if log space > set size?

---
In reply to: [921140320](https://github.com/microsoft/do-client/pull/76#issuecomment-921140320) [](http://example.com/codeflow?ancestors=921140320,921139047,921139047)

shishirb-MSFT:
Not worth having that code to perform log deletion, etc. The plugin can be run concurrently in multiple processes so we'll have to deal with that as well.

---
In reply to: [921141125](https://github.com/microsoft/do-client/pull/76#issuecomment-921141125) [](http://example.com/codeflow?ancestors=921141125,921139047)

plugins/linux-apt/main.cpp
```diff
@@ -315,22 +316,23 @@ int main(int argc, char** argv)
             printf("Unknown option: %s\n", argv[1]);
             return 1;
         }
+#endif // DEBUG
```
jimson-msft, line 319: nit: in the future we could add these to a few unit tests, also thinking it would be a good place to add tests that download e2e (instruct apt to download package -> verify package download succeeds -> test cleanup uninstalls package & clears apt cache)

jimson-msft (approved):
:shipit:

plugins/linux-apt/main.cpp
shishirb-MSFT: agreed

PR #77, opened 2021-09-16
Fix windows build break

jimson-msft:


sdk-cpp/src/do_download_property.cpp
```diff
@@ -2,6 +2,7 @@
 #if defined(DO_INTERFACE_COM)
 #include "do_download_property.h"
 
+#include <cassert>
```
shishirb-MSFT, line 5: >cassert [](http://example.com/codeflow?start=10&length=7)

Strange I didn't see this failure earlier this morning.

shishirb-MSFT (approved):
:shipit:

PR #78, opened 2021-09-17
ARM pipelines: Fix version parameter name

shishirb-MSFT:


azure-pipelines/build/linux/du/dopapt-docker.yml
```diff
@@ -26,17 +26,17 @@ jobs:
 - template: templates/do-docker-jobs.yml
```
shishirb-MSFT, line 26: >docker [](http://example.com/codeflow?start=25&length=6)

Plugin build is failing with 
CMake Error at /usr/share/cmake-3.7/Modules/CMakeFindDependencyMacro.cmake:35 (message):
  Invalid arguments to find_dependency

Any clues?

jimson-msft: That is bizarre. Let me spin up a container and see what's going on internally

jimson-msft: So it looks like the deliveryoptimization_sdk-config.cmake is exported with the line "find_package(Boost COMPONENTS filesystem system REQUIRED)"

The 'components' argument is not supported until a version of cmake that is later than whats natively available on deb9.
Not sure how we didn't hit this issue before though, but I kind of do vaguely recall seeing this

jimson-msft: also: system no long required for sdk builds I think

jimson-msft: I recall we did hit this issue before, replacing find_dependency with find_package was the solution.
There must have been an issue with merge conflict resolution leading to us goign back to somehow using find_dependency. Let me find it.

shishirb-MSFT: Yeah I wonder how we did not hit this before. The public repo has find_dependency since the public preview time.

shishirb-MSFT: Previous docker image had us building cmake from source (v3.10). Fixed the sdk cmake config file. Plugin arm build running now.

jimson-msft (approved):

PR #80, opened 2021-09-21
[Draft] Remove internal exceptions & Expose non-throwing apis

jimson-msft:
- Enable building SDK without exceptions 
- Remove internal exceptions except for:
1. There are still internal exceptions in our utility classes for the rest interface side of things. Looking for feedback here.
2. download_property class is quite awkward if we remove exceptions, it feels awkward to have an Init/Clear method for it since it's a thin wrapper around VARIANT or another polymorphic type for the rest interface. Looking for feedback here. 
- Tests run succesfully across both windows/linux

### DIFF
--- a/sdk-cpp/include/do_exceptions.h
+++ b/sdk-cpp/include/do_exceptions.h
@@ -1,21 +1,46 @@
 #ifndef _DELIVERY_OPTIMIZATION_DO_EXCEPTIONS_H
 #define _DELIVERY_OPTIMIZATION_DO_EXCEPTIONS_H
 
+#if (!DO_DISABLE_EXCEPTIONS)
+
 #include <exception>
 #include <stdint.h>
 #include <system_error>
 
-#if defined(DO_INTERFACE_COM)
-#include <winerror.h>   // FAILED macro
-#endif
+#endif // !DO_DISABLE_EXCEPTIONS
 
 namespace microsoft
 {
 namespace deliveryoptimization
 {
 
+// Error Macros
+#ifndef RETURN_HR_IF_FAILED
+#define RETURN_HR_IF_FAILED(hr)  {  \
+    int32_t __hr = (hr);            \
+    if(FAILED(__hr)) return __hr;   \
+}
+#endif
+
+#ifndef FACILITY_DELIVERY_OPTIMIZATION 
+#define FACILITY_DELIVERY_OPTIMIZATION   208
+#endif
+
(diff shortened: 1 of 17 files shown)


jimson-msft:
@shishirb-MSFT Looking for feedback on exception removal in do_download_property class. We use RAII for handling things like VariantInit and VariantClear here, so I could expose a Init/Clear method do_download_property as well, but that feels a little awkward to have as a top-level api.

There are also still internal exceptions within the rest/util files. I can look to remove our own exceptions there and wrap every 3p library call in a try/catch. Let me know if that sounds good.

sdk-cpp/include/do_exceptions.h
```diff
@@ -1,21 +1,46 @@
...
+
+#define HRESULT_FROM_SYSTEM_ERROR(x) ((int32_t)(x) <= 0 ? ((int32_t)(x)) : ((int32_t) (((x) & 0x0000FFFF) | (FACILITY_DELIVERY_OPTIMIZATION << 16) | 0x80000000)))
+
+#ifndef FAILED
+#define FAILED(hr) (((int32_t)(hr)) < 0)
+#endif
+
+#ifndef SUCCEEDED
+#define SUCCEEDED(hr) (((int32_t)(hr)) >= 0)
+#endif
```
jimson-msft, line 37: Without exceptions, need a way to identify failing error codes. Which is a little awkward because we draw from a variety of error codes. (HRESULTS, std::error_code, our own error code enum, etc)

In the case of std::error_code, these values are defined as positive integers, and so error checking macro fail/succeeded wouldn't catch these failures appropriatley.

I introduced HRESULT_FROM_SYSTEM_ERROR as a result, and swapped out the facility code with our own, but I'm not sure if this is the best approach.

sdk-cpp/include/do_exceptions.h
```diff
@@ -1,21 +1,46 @@
...
+    if(FAILED(__hr)) return __hr;   \
+}
+#endif
+
+#ifndef FACILITY_DELIVERY_OPTIMIZATION 
+#define FACILITY_DELIVERY_OPTIMIZATION   208
+#endif
+
+#define HRESULT_FROM_SYSTEM_ERROR(x) ((int32_t)(x) <= 0 ? ((int32_t)(x)) : ((int32_t) (((x) & 0x0000FFFF) | (FACILITY_DELIVERY_OPTIMIZATION << 16) | 0x80000000)))
+
```
jimson-msft, line 30: Actually, now I'm not so sure that this is a good idea. It'll kind of morph the error code and become harder to understand what it corresponds to.

I think what we can do is define all error codes in the errc enum below.

jimson-msft:
> }

I think I will go with the approach of having init() and clear() here to remove exceptions from this class, also will make it available on the rest side.

---
Refers to: sdk-cpp/include/do_download_property.h:95 in 893d1d5. [](commit_id = 893d1d54f63e609958c7361840b07336ba14a324, deletion_comment = False)

PR #81, opened 2021-09-23
Make header files platform agnostic and remove all internal exceptions

jimson-msft:


CMakeLists.txt
```diff
@@ -61,6 +61,10 @@ endif()
 # PIC (Position Independent Code) ensures .a files can be linked to executables that have PIE enabled
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+if (DO_DISABLE_EXCEPTIONS)
+    #TODO(jimson): Set cmake flags for disabling exceptions and rtti, there is no cmake setting for this, so we must set the compiler-specific flag
+endif()
```
jimson-msft, line 66: This flag is compiler dependent - not sure if we should explicitly set it here (might help us actually verify our code will compile against -fno-except) but also we dont have a way to know if this is true for all compiler/toolchain combos we need.

build/build.py
```diff
@@ -81,6 +81,12 @@ def __init__(self):
             help='Override default build output directory'
         )
 
+        '''SDK only'''
+        self.parser.add_argument(
+            '--disable-exceptions', dest='disable_exceptions', action='store_true',
+            help='Builds libdeliveryoptimization without compiling exception throwing code into the binary'
+        )
+
```
jimson-msft, line 89: I went with this convention because this allows for opting out of exceptions, and it doesn't break the current interface.

By default, both the throwing and non-throwing public apis will be available

sdk-cpp/include/do_config.h
```diff
@@ -1,9 +1,10 @@
 #ifndef _DELIVERY_OPTIMIZATION_DO_CONFIG_H
 #define _DELIVERY_OPTIMIZATION_DO_CONFIG_H
 
-//TODO(jimson): Callers may not have defined these compile definitions, as a result their builds may fail if the definition is not set when using the SDK
-//Look into removing platform specific header files from the SDK installation
-#if defined(DO_CLIENT_AGENT)
+/* User Note:
+This file exposes apis for helping set an iot_connection_string for the DO agent, so that an application written in c can supply a Microsoft Connected Cache device's hostname
+While the apis will compile for all platforms, they only serve a purpose for the DeliveryOptimization Agent on linux devices, all other usage will fail and return e_not_impl
+*/
```
jimson-msft, line 7: I went with this because I wanted a way to differentiate technical comments on implementation details from usability comments

sdk-cpp/include/do_download.h
```diff
@@ -26,49 +24,43 @@ class download
...
     void start_and_wait_until_completion(const std::atomic_bool& isCancelled, std::chrono::seconds timeoutSecs = std::chrono::hours(24));
-
     static void download_url_to_path(const std::string& uri, const std::string& downloadFilePath, std::chrono::seconds timeoutSecs = std::chrono::hours(24));
     static void download_url_to_path(const std::string& uri, const std::string& downloadFilePath, const std::atomic_bool& isCancelled, std::chrono::seconds timeoutSecs = std::chrono::hours(24));
 
-#if defined(DO_INTERFACE_COM)
-    /* Sets a download property for the current download object
-    This function attempts to set a download property using the caller-supplied value. This method can throw
-    ~~~~
+    /*
```
jimson-msft, line 40: TODO: add "User Note" here

sdk-cpp/include/do_download_property.h
```diff
@@ -42,56 +46,58 @@ enum class download_property
...
     http_custom_auth_headers,       // std::string
 };
 
 class download_property_value
 {
 
+// CDownloadImpl is declared as a friend class because it needs to access the platform-specific native value for download_property_value
+// The type of the native value is defined in CDownloadPropertyValueInternal, because DO header files are platform agnostic
+// This is so any user of the SDK does not have to worry about supplying platform specific compile definitions to use the SDK
+friend class details::CDownloadImpl;
```
jimson-msft, line 61: I'm entirely not sure if the way I implemented this is the best way or if this is a proper usage of friend class.

The other option I thought about was just in time initialization of the native type inside download_property_impl and just scrubbing this class altogether. I couldn't see why a user would want to hold onto the download property value.

If we want to go with that approach, I can just expose a bunch of overloads to SetProperty() accepting these various types, and then create the VARIANT internally in download_impl

sdk-cpp/include/do_download_property.h
```diff
@@ -42,56 +46,58 @@ enum class download_property
...
+    */
     explicit download_property_value(const std::string& val);
-    explicit download_property_value(UINT val);
-    explicit download_property_value(UINT64 val);
+    explicit download_property_value(uint32_t val);
+    explicit download_property_value(uint64_t val);
     explicit download_property_value(bool val);
     explicit download_property_value(std::vector<unsigned char>& val); // DODownloadProperty_SecurityContext
 
     explicit download_property_value(const status_callback_t& val);
```
jimson-msft, line 76: This was the other iffy thing to me - it feels weird to compile the SDK in a way such that these constructors will have different behavior. But it is definitely a working solution - just not sure if it's the best one.

sdk-cpp/include/do_exceptions.h
```diff
@@ -1,21 +1,45 @@
...
+#define DO_ERROR_FROM_SYSTEM_ERROR(x) (int32_t)(0xC0000000 | (0xD0 << 16) | ((int32_t)(x) & 0x0000FFFF))
+#define DO_ERROR_FROM_STD_ERROR(x) ((int32_t)(x) <= 0 ? ((int32_t)(x)) : ((int32_t) (((int32_t)(x) & 0x0000FFFF) | (FACILITY_DELIVERY_OPTIMIZATION << 16) | 0x80000000)))
+
+#ifndef FAILED
+#define FAILED(res) (((int32_t)(res)) < 0)
+#endif
+
+#ifndef SUCCEEDED
+#define SUCCEEDED(res) (((int32_t)(res)) >= 0)
+#endif
```
jimson-msft, line 37: I used RES here, but I think I could also replace it with 'e' for brevity, as well as all int32_t locals

CMakeLists.txt
```diff
@@ -61,6 +61,10 @@ endif()
 # PIC (Position Independent Code) ensures .a files can be linked to executables that have PIE enabled
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+if (DO_DISABLE_EXCEPTIONS)
```
shishirb-MSFT, line 64: >DO_DISABLE_EXCEPTIONS [](http://example.com/codeflow?start=4&length=21)

Don't we have to add this as an option above like DO_BUILD_TESTS?

jimson-msft: Oh, we should do what we did for DO_BUILD_TESTS above, but it's not actually required to set it in the build. I think the above is to just set it as a cmake option

shishirb-MSFT, line 64: >DO_DISABLE_EXCEPTIONS [](http://example.com/codeflow?start=4&length=21)

How about we go with opt-in for exceptions? DO_ENABLE_EXCEPTIONS. WIL does the same.

The disable word introduces a double negative when writing !DO_DISABLE_EXCEPTIONS which I'd like to avoid.

CMakeLists.txt
shishirb-MSFT: We could set the correct flag based on the compiler but I agree this isn't blocking.

CMakeLists.txt
jimson-msft: Yeah I had considered this but the problem with opt-in is that it doesn't remove the throwing code.
There is internal code which may still utilize try/catch or throw keywords that need to be disabled by an application compiling with -fno-except

sdk-cpp/include/do_download_property.h
shishirb-MSFT: The standard convention here is to provide make() methods and a default constructor that is private.

int32_t make(const std::string& val);
int32_t make(uint32_t val);
...

CMakeLists.txt
jimson-msft: Yes, I also need to modify the comment because it reads a little confusingly. I'll modify it to say 
"There is no cmake option for disabling exceptions/RTTI, so we must set a compiler-specific flag if we want to enforce disabling exceptions"

sdk-cpp/include/do_download_property.h
jimson-msft: But isn't this a little awkward for consumers who will want to use a throwing interface? They might expact constructors utilizing RAII and throwing on failure. Should we expect people using both the throwing/non-throwing interface to use make() methods?

shishirb-MSFT:
>     _download = std::make_shared<msdod::CDownloadImpl>(uri, downloadFilePath);

Question: If the compiler has exceptions turned off, what happens to memory allocation failures? Can you find out what chromium does?

---
Refers to: sdk-cpp/src/do_download.cpp:22 in 88a698b. [](commit_id = 88a698b07aaafc7644eaba5b8835826254ef1d6b, deletion_comment = False)

sdk-cpp/include/do_exceptions.h
```diff
@@ -1,21 +1,45 @@
...
+
 enum class errc : int32_t
 {
+    s_ok                        = 0,
```
shishirb-MSFT, line 41: >s_ok [](http://example.com/codeflow?start=4&length=4)

errc enum should not include a success code.

Since we have a non-throwing interface now, we should invest in a msdo::error_code class modeled on std::error_code or reuse std::error_code by supplying our own error_category object (I believe cpprest did something like this that you can borrow).

sdk-cpp/include/do_exceptions.h
```diff
@@ -1,21 +1,45 @@
...
 
+//These error macros are defined behind guards so that they do not redefine the macro for for platforms/toolchains in which they already exist
+#ifndef RETURN_RES_IF_FAILED
+#define RETURN_RES_IF_FAILED(res)  {  \
+    int32_t __res = (res);            \
+    if(FAILED(__res)) return __res;   \
+}
+#endif
+
+#ifndef FACILITY_DELIVERY_OPTIMIZATION
```
shishirb-MSFT, line 24: >FACILITY_DELIVERY_OPTIMIZATION [](http://example.com/codeflow?start=8&length=30)

Can you pull out these definitions from client-lite/src/include/hresult_helpers.h into a new header in common/inc/basic_error_helpers.h?

sdk-cpp/src/internal/download_interface.h
```diff
@@ -3,9 +3,9 @@
 
 #include "do_download_status.h"
 
-#if defined(DO_INTERFACE_COM)
 #include "do_download_property.h"
 
+#if defined(DO_INTERFACE_COM)
```
shishirb-MSFT, line 8: >DO_INTERFACE_COM [](http://example.com/codeflow?start=12&length=16)

The SetCallback method is no longer under ifdef, so this one shouldn't be either, no?

sdk-cpp/src/internal/do_config_internal.h
```diff
@@ -0,0 +1,12 @@
+#ifndef _DELIVERY_OPTIMIZATION_DO_CONFIG_INTERNAL_H
```
shishirb-MSFT, line 1: What is the need to create this internal config API? Is it because it cannot be converted to not use try/catch?

sdk-cpp/src/internal/do_download_property_internal.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "do_download_property.h"
+
+namespace microsoft
+{
+namespace deliveryoptimization
+{
+namespace details
+{
+class CDownloadPropertyValueInternal
```
shishirb-MSFT, line 22: >CDownloadPropertyValueInternal [](http://example.com/codeflow?start=6&length=30)

We don't need a separate internal class when the native_type is controlled by a compile-time constant. Or am I missing something?

jimson-msft:
>     _download = std::make_shared<msdod::CDownloadImpl>(uri, downloadFilePath);

Based on what I'd read aboutapps in general that compile with exceptions disabled - they crash on memory allocation failures.

---
In reply to: [926062947](https://github.com/microsoft/do-client/pull/81#issuecomment-926062947) [](http://example.com/codeflow?ancestors=926062947,926062947)

---
Refers to: sdk-cpp/src/do_download.cpp:22 in 88a698b. [](commit_id = 88a698b07aaafc7644eaba5b8835826254ef1d6b, deletion_comment = False)

sdk-cpp/src/internal/com/download_impl.cpp
```diff
@@ -186,109 +209,134 @@ class DOStatusCallback :
         }
         return S_OK;
     }
+#else
+    // If an application builds the sdk from source and has toggled DO_DISABLE_EXCEPTIONS, it would be hypocritical for their callback to throw
```
shishirb-MSFT, line 213: >hypocritical [](http://example.com/codeflow?start=103&length=12)

LOL

sdk-cpp/include/do_exceptions.h
jimson-msft: Good point, yeah I'll look into creating a custom error code class. Even boost has boost::system::error_code

sdk-cpp/include/do_exceptions.h
jimson-msft: Sounds good, that file is pulled from WIl right? So we're good to use it over

sdk-cpp/src/internal/download_interface.h
jimson-msft: Right, will remove

sdk-cpp/src/internal/do_config_internal.h
jimson-msft: The public header needs to be platform agnostic, particularly for cases where we build & install the sdk as a package. Sorry - this PR is confusing because I added whats really two changes into one PR.

sdk-cpp/src/internal/do_download_property_internal.h
jimson-msft: Isn't the using directive already a compile time constant? (I may be misunderstanding here). But yes the reason for this is to abstract away the native_value type so that the header contains no traces of
(DO_INTERFACE_COM) or (DO_INTERFACE_REST)

sdk-cpp/src/internal/com/download_impl.cpp
jimson-msft: 😂

sdk-cpp/src/internal/com/download_impl.cpp
```diff
@@ -186,109 +209,134 @@ class DOStatusCallback :
...
-    }
     _spDownload = std::move(spDownload);
+    return static_cast<int32_t>(errc::s_ok);
 }
 
 // Support only full file downloads for now
-void CDownloadImpl::Start()
+int32_t CDownloadImpl::Start() noexcept
 {
+    RETURN_RES_IF_FAILED(_codeInit);
```
shishirb-MSFT, line 258: >RETURN_RES_IF_FAILED(_codeInit); [](http://example.com/codeflow?start=4&length=32)

Yeah this type of thing is a code smell. I'll describe the cleaner option in the do_download_property.h comment.

sdk-cpp/include/do_download_property.h
shishirb-MSFT: We can offer both throwing and non throwing versions of make methods too.

Think of it as this way - constructors should no longer do anything that can fail (except allocating memory/other objects of course). If anything more is required, for example other initialization steps as in the CDownloadImpl example, then we should make the constructors private and offer 'builder' methods (the make methods) which can have both throwing and non-throwing versions.

sdk-cpp/src/internal/com/download_impl.cpp
```diff
@@ -186,109 +209,134 @@ class DOStatusCallback :
...
+    Microsoft::WRL::ComPtr<DOStatusCallback> spCallback;
+    RETURN_RES_IF_FAILED(MakeAndInitialize<DOStatusCallback>(&spCallback, callback, download));
+
+    VARIANT vtCallback;
+    VariantInit(&vtCallback);
+    V_VT(&vtCallback) = VT_UNKNOWN;
+    V_UNKNOWN(&vtCallback) = spCallback.Get();
+    spCallback.Get()->AddRef();
+    DODownloadProperty prop;
+    RETURN_RES_IF_FAILED(ConvertToComProperty(msdo::download_property::callback_interface, prop));
```
shishirb-MSFT, line 323: >RETURN_RES_IF_FAILED [](http://example.com/codeflow?start=4&length=20)

This shouldn't fail, yes? Either convert to assert (with ifdef DEBUG, else) or just don't check the return value.

shishirb-MSFT: Also remember that when we return fast like this we should perform cleanup like VariantClear before returning.

sdk-cpp/src/internal/com/download_impl.cpp
```diff
@@ -186,109 +209,134 @@ class DOStatusCallback :
...
 
-void CDownloadImpl::_SetPropertyHelper(IDODownload& download, msdo::download_property key, const msdo::download_property_value& val)
+int32_t CDownloadImpl::_SetPropertyHelper(IDODownload& download, msdo::download_property key, const msdo::download_property_value& val) noexcept
 {
-    DODownloadProperty prop = ConvertToComProperty(key);
+    DODownloadProperty prop;
+    RETURN_RES_IF_FAILED(ConvertToComProperty(key, prop));
 
+    std::shared_ptr<CDownloadPropertyValueInternal> prop_val = val._val;
     // COM API SetProperty() is not declared with const reference for variant argument so need to const_cast
```
shishirb-MSFT, line 335: >const [](http://example.com/codeflow?start=50&length=5)

It is const now so should be able to remove the cast. (can the done later too).

sdk-cpp/src/internal/com/download_impl.cpp
```diff
@@ -186,109 +209,134 @@ class DOStatusCallback :
...
 }
 
-void CDownloadImpl::_SetPropertyHelper(IDODownload& download, msdo::download_property key, const msdo::download_property_value& val)
+int32_t CDownloadImpl::_SetPropertyHelper(IDODownload& download, msdo::download_property key, const msdo::download_property_value& val) noexcept
 {
-    DODownloadProperty prop = ConvertToComProperty(key);
+    DODownloadProperty prop;
+    RETURN_RES_IF_FAILED(ConvertToComProperty(key, prop));
 
+    std::shared_ptr<CDownloadPropertyValueInternal> prop_val = val._val;
```
shishirb-MSFT, line 334: >prop_val [](http://example.com/codeflow?start=52&length=8)

Why make a copy and not use val._val->native_value() ?

sdk-cpp/src/internal/com/do_download_property_internal.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+int32_t CDownloadPropertyValueInternal::init_code() const noexcept
+{
+    return _initCode;
+}
+
+CDownloadPropertyValueInternal::~CDownloadPropertyValueInternal()
+{
+#ifdef DEBUG
+    //TODO(jimson): Variant clear fails with DISP_E_BADVARTYPE when used for callbacks, so the assertion will terminate the application
```
shishirb-MSFT, line 93: >DISP_E_BADVARTYPE [](http://example.com/codeflow?start=45&length=17)

This is most likely because VariantInit is not done. See line 82 above.

sdk-cpp/src/internal/rest/util/do_http_client.cpp
```diff
@@ -70,22 +69,27 @@ CHttpClient& CHttpClient::GetInstance()
...
     }
 
     std::unique_lock<std::mutex> lock(_mutex);
     _httpClientImpl = std::move(spImpl);
+
+    return static_cast<int32_t>(microsoft::deliveryoptimization::errc::s_ok);
 }
 
-boost::property_tree::ptree CHttpClient::SendRequest(HttpRequest::Method method, const std::string& url, bool retry)
+int32_t CHttpClient::SendRequest(boost::property_tree::ptree& tree, HttpRequest::Method method, const std::string& url, bool retry)
```
shishirb-MSFT, line 92: >boost::property_tree::ptree& tree [](http://example.com/codeflow?start=33&length=33)

Out params should come after the in params.

sdk-cpp/src/internal/rest/util/do_http_client.cpp
```diff
@@ -100,19 +104,22 @@ boost::property_tree::ptree CHttpClient::SendRequest(HttpRequest::Method method,
...
     }
 
     if (responseStatusCode != 200)
     {
         auto agentErrorCode = responseBodyJson.get_optional<int32_t>("ErrorCode");
-        ThrowException(agentErrorCode ? *agentErrorCode : -1);
+        if (agentErrorCode)
+        {
+            return agentErrorCode.get();
+        }
```
shishirb-MSFT, line 119: >} [](http://example.com/codeflow?start=8&length=1)

else
{
    return static_cast<int32_t>(msdo::errc::unexpected);
}

sdk-cpp/src/internal/rest/util/do_http_client.cpp
```diff
@@ -70,22 +69,27 @@ CHttpClient& CHttpClient::GetInstance()
...
 }
 
-void CHttpClient::_InitializeDOConnection(bool launchClientFirst)
+int32_t CHttpClient::_InitializeDOConnection(bool launchClientFirst)
 {
-    const auto port = std::strtoul(CPortFinder::GetDOPort(launchClientFirst).data(), nullptr, 10);
+    std::string portStr;
+    int32_t res = CPortFinder::GetDOPort(portStr, launchClientFirst);
+    const auto port = std::strtoul(portStr.data(), nullptr, 10);
+    RETURN_RES_IF_FAILED(res);
```
shishirb-MSFT, line 77: >RETURN_RES_IF_FAILED(res); [](http://example.com/codeflow?start=4&length=26)

Return before calling strtoul.

sdk-cpp/src/internal/rest/util/do_port_finder.cpp
```diff
@@ -7,7 +7,6 @@
 
 #include <boost/filesystem.hpp>
 
-#include "do_exceptions_internal.h"
 #include "do_exceptions.h"
```
shishirb-MSFT, line 10: Need this?

shishirb-MSFT: Oh for the error codes.

CMakeLists.txt
shishirb-MSFT: I don't follow. Our code should not use try/catch unless DO_ENABLE_EXCEPTIONS is defined. If it is about 3rd party libs, then it applies to DO_DISABLE_EXCEPTIONS as well.

shishirb-MSFT:
>     _download = std::make_shared<msdod::CDownloadImpl>(uri, downloadFilePath);

They crash on using a nullptr or unhandled bad_alloc exception? Specifically, does chromium replace the 'new' operator with a non-throwing version globally?

---
In reply to: [926078792](https://github.com/microsoft/do-client/pull/81#issuecomment-926078792) [](http://example.com/codeflow?ancestors=926078792,926062947)

---
Refers to: sdk-cpp/src/do_download.cpp:22 in 88a698b. [](commit_id = 88a698b07aaafc7644eaba5b8835826254ef1d6b, deletion_comment = False)

sdk-cpp/src/internal/do_config_internal.h
shishirb-MSFT: Are you going to separate it as discussed? To release a new version for DU...

CMakeLists.txt
jimson-msft: We can do that, but we'd need to bump the major version of the SDK after doing so, because a consumer would now need to set that compile option. With the current implementation, we don't necessarily hae to bump the majorversion since it's only opt-in to disable

jimson-msft: I'm going to leave it as is for now, and I can check in a subsequent change using DO_ENABLE_EXCEPTIONS and bumping the majorversion when ready

sdk-cpp/src/internal/com/do_download_property_internal.cpp
jimson-msft: Ah yes, good catch

jimson-msft:
>     _download = std::make_shared<msdod::CDownloadImpl>(uri, downloadFilePath);

Neha shared this: 
[https://google.github.io/styleguide/cppguide.html](https://google.github.io/styleguide/cppguide.html)

Note in many cases the only possible cause for an exception is allocation failure (we believe move constructors should not throw except due to allocation failure), and there are many applications where it’s appropriate to treat memory exhaustion as a fatal error rather than an exceptional condition that your program should attempt to recover from. Even for other potential failures you should prioritize interface simplicity over supporting all possible exception throwing scenarios: instead of writing a complicated noexcept clause that depends on whether a hash function can throw, for example, simply document that your component doesn’t support hash functions throwing and make it unconditionally noexcept.

---
In reply to: [926371725](https://github.com/microsoft/do-client/pull/81#issuecomment-926371725) [](http://example.com/codeflow?ancestors=926371725,926062947)

---
Refers to: sdk-cpp/src/do_download.cpp:22 in 88a698b. [](commit_id = 88a698b07aaafc7644eaba5b8835826254ef1d6b, deletion_comment = False)

sdk-cpp/src/internal/do_config_internal.h
jimson-msft: Yes I can look into splitting this PR into two. I'll resolve all current comments, and re-open two separate PR's, the first change will be the platform agnostic headers (which we can set as the release commit if things look good), and then the exceptions change.

jimson-msft:
@device is added to the review. #Closed

CMakeLists.txt
shishirb-MSFT: Note that only API changes require major version increment, not changes in the way it builds. Although one could argue that adding noexcept is a change to the API.

We are in preview phase so we should take advantage of it for such changes (thinking about DU here). Moreover this is the first time Edge will consume this, so let's make this change now.

sdk-cpp/include/do_exceptions.h
shishirb-MSFT: This one can come as a later PR into the feature branch (if you take my suggestion of completing this change as piecemeal PRs).

CMakeLists.txt
shishirb-MSFT: Ok I'd still expose it as an option so anyone can see at a glance what configurations we support.

jimson-msft: Yes, will do. I had mentioned that after splitting this PR, we can easily make the exception change and choose to increment the majorversion (or leave as is). I'll make the change in a subsequent PR

PR #82, opened 2021-09-27
Add copyright notices to all files

jimson-msft:
- Following Microsoft CELA guidance for copyright notices across all source files.
- Files borrowed from cpprest retain cpprest license (it is up to modifier if we want to add our own contribution, but we must retain the original source files license - so the cpprest header notices are taken directly from the projects github repo)


build/check_binary_size.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
```
shishirb-MSFT, line 1: This file and bootstrap.sh are dupliated in build/scripts. Merge issue?

shishirb-MSFT:
Thanks for adding the cpprest header. Also, just curious - did you use any tool to automatically add the header to all files?

cmknox:
Thanks @jimson-msft for adding the header to these files!

build/check_binary_size.sh
jimson-msft: Yea these files must have been retained in the previous merge, but they should be deleted now.. I'll remove in this PR

shishirb-MSFT (approved):
:shipit:

PR #83, opened 2021-09-27
Make headers platform agnostic

jimson-msft:


jimson-msft:
> 

Oops, need to delete.
Was originally just trying ot use this file to obfuscate the compile flag for the variant type, but its no good to include a private header in the public headers

---
Refers to: sdk-cpp/src/internal/do_variant_native_type.h:22 in 6b9e1ad. [](commit_id = 6b9e1adcd8c2d881760c2ab371c7e3df362ac6ab, deletion_comment = False)

sdk-cpp/src/internal/do_download_property_internal.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "do_download_property.h"
+
+namespace microsoft
+{
+namespace deliveryoptimization
+{
+namespace details
+{
+class CDownloadPropertyValueInternal
```
jimson-msft, line 22: re:
"We don't need a separate internal class when the native_type is controlled by a compile-time constant. Or am I missing something?"

I'm not sure if there's a way to do the above. Need some way to hide the platform specific implementation. 

Considered forward declaring the typdef, but it also requires the type to be fwd declared.

jimson-msft: One thing I had considered was just removing the do_download_property class altogether, and just providing overloads for msdo::download::set_property() to accept all the different types, and then to convert the type to the respective native type in the download_impl class.
I forget why we wanted to let the caller manage a download_property class, because once it's constructed it's immutable in the current implementation.

shishirb-MSFT: Providing overloads for msdo::download::set_property() --
sounds good. See how it works out in terms of coding and API clarity.

But what about get_property? The caller will not know which overload to call.

sdk-cpp/include/do_config.h
```diff
@@ -4,36 +4,35 @@
 #ifndef _DELIVERY_OPTIMIZATION_DO_CONFIG_H
 #define _DELIVERY_OPTIMIZATION_DO_CONFIG_H
 
-//TODO(jimson): Callers may not have defined these compile definitions, as a result their builds may fail if the definition is not set when using the SDK
-//Look into removing platform specific header files from the SDK installation
-#if defined(DO_CLIENT_AGENT)
+/*
+This file exposes apis for helping set an iot_connection_string for the DO agent, so that an application written in c can supply a Microsoft Connected Cache device's hostname
```
shishirb-MSFT, line 8: >c [](http://example.com/codeflow?start=116&length=1)

uppercase C

shishirb-MSFT: Note that this API can be called from C++ also.

sdk-cpp/include/do_download_property.h
```diff
@@ -45,56 +49,48 @@ enum class download_property
     integrity_check_info,           // std::string
     integrity_check_mandatory,      // boolean
     total_size_bytes,               // uint64
-    // Fe and beyond
+
+    // For the COM interface, the following properties are available only in Windows 21H1 and beyond
```
shishirb-MSFT, line 53: >21H1 [](http://example.com/codeflow?start=85&length=4)

21H1 is still the Vibranium build. Let's go with the build number for Cobalt here (22000) since that is the first desktop release with support for these properties.

sdk-cpp/src/internal/do_download_property_internal.h
jimson-msft: I agree the pimpl idiom is not great just so we can abstract away the underlying native type that we use to make calls to our client, but I'm not too sure on which I prefer between what's in this PR and then using multiple overloads for set_property()... thinking

sdk-cpp/include/do_download_property.h
jimson-msft: modified comment

shishirb-MSFT (approved):
:shipit:

sdk-cpp/include/do_download_property.h
```diff
@@ -45,56 +49,48 @@ enum class download_property
...
 
     void as(status_callback_t& val) const;
     void as(std::vector<unsigned char>& val) const;
 
-    const native_type& native_value() const;
-
 private:
-    native_type _var;
-    status_callback_t _callback;
+    std::unique_ptr<details::CDownloadPropertyValueInternal> _val;
```
jimson-msft, line 91: Actually, not so straightforward to convert to unique_ptr, since it's a unique_ptr class must define a move constructor which forwards _val into the copy, and delete the copy constructor.
Since we this class is used as a return value internally - we must provide do the above.
But at the same time we cannot delete an incomplete type of CDownloadPropertyValImpl stored in the unique ptr if we try to delete the copy constructor

jimson-msft: Anyways, because this may get refactored in the future, going to use std::shared_ptr

shishirb-MSFT: Sounds good.

PR #84, opened 2021-09-28
Prepare to release v0.8.0

shishirb-MSFT:
* agent, sdk: 0.6.0
* plugin-apt: 0.4.0

### DIFF
--- a/azure-pipelines/publishing/github-release.yml
+++ b/azure-pipelines/publishing/github-release.yml
@@ -6,12 +6,15 @@ trigger:
 pr:
 - none
 
-pool: 1es_hosted_pool_ubuntu_1804
+pool:
+  name: 1es_hosted_pool_ubuntu_1804
+  demands: ImageOverride -equals do-adu-build-$(imageVersion)
 
 variables:
   Release.Version: $(release_version)
   Release.PreviousVersion: $(previous_release_version) # Select the previous version, used to generate changelog
   Release.Title: $(title) # title of the release, following <Release.Version> (YYYY-MM-DD) format i.e. v0.6.0 (2021-03-02)
+  imageVersion: '0.8.0'
 
 parameters:
   - name: SkipPublishing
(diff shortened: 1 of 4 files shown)


shishirb-MSFT:
I expect some changes to come after testing with the DU yocto build. Wait for further iterations.

shishirb-MSFT:
Tested ubuntu 1804 packages for both the fresh install and upgrade scenario. Any further changes will be done as patch releases.

---
In reply to: [929331828](https://github.com/microsoft/do-client/pull/84#issuecomment-929331828) [](http://example.com/codeflow?ancestors=929331828)

jimson-msft:
Not going to wait for the platform agnostic header change? Keep in mind this will break DU builds unless we notify them that they need to add the appropriate DO_INTERFACE_REST and DO_CLIENT_AGENT flags (may also break mariner builds when they adopt)

shishirb-MSFT:
Oops, forgot about it. I'll wait.

---
In reply to: [929538113](https://github.com/microsoft/do-client/pull/84#issuecomment-929538113) [](http://example.com/codeflow?ancestors=929538113,929538113)

jimson-msft:
> Oops, forgot about it. I'll wait.
> 
> In reply to: [929538113](https://github.com/microsoft/do-client/pull/84#issuecomment-929538113)

Just checked the other PR in, this change looks good feel free to check in. I do notice reproducible failure with the plugin pipeline where the dpkg lock is already being held:
https://deliveryoptimization.visualstudio.com/client/_build/results?buildId=3423&view=logs&j=9d538be6-44f8-50c3-7930-4e7f88a0f6b2&t=4609d989-b212-5a4b-e936-e5e86def8716

Not sure what the cause is, it's happening on multiple agent machines as well

jimson-msft (approved):
:shipit:

shishirb-MSFT:
Yeah I saw that and wasn't sure why it is happening only for the version updates. I'll dig deeper.

---
In reply to: [929730622](https://github.com/microsoft/do-client/pull/84#issuecomment-929730622) [](http://example.com/codeflow?ancestors=929730622,929730622)

shishirb-MSFT:
Not seeing it anymore. Must be a transient issue with the VM.

---
In reply to: [929827600](https://github.com/microsoft/do-client/pull/84#issuecomment-929827600) [](http://example.com/codeflow?ancestors=929827600,929730622,929730622)

PR #86, opened 2021-10-01
Increment versions for develop

shishirb-MSFT:
* Increment post-v0.8.0 release.
* The idea is to be able to always differentiate between released and development versions.

### DIFF
--- a/client-lite/CMakeLists.txt
+++ b/client-lite/CMakeLists.txt
@@ -5,7 +5,7 @@ if (NOT DOSVC_BIN_NAME)
     message (FATAL_ERROR "Agent daemon name not defined")
 endif ()
 
-project (${DOSVC_BIN_NAME} VERSION 0.6.0)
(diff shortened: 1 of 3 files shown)


jimson-msft (approved):

jimson-msft:
The only thought I have here is it's hard to know if the subsequent release is minorversion or majorversion impacting. For example the exceptions changes introduced *might* warrant a majorversion increase, but they might not given that the existing interfaces shouldnt change

shishirb-MSFT:
Yeah we have some flexibility now, while in preview phase. If the next release needs major version increase, then we can and will do it just before releasing. The challenge is once we are stable, whether to know if the next release will contain only bug fixes or feature also (patch version vs. minor version change).

---
In reply to: [932460812](https://github.com/microsoft/do-client/pull/86#issuecomment-932460812) [](http://example.com/codeflow?ancestors=932460812,932460812)

PR #88, opened 2021-10-02
Error Code Interface & Partially Remove Internal Exceptions

jimson-msft:
- Provide error code returning interface
- Strip internal exceptions for win10
- For rest, internal classes may call 3p libraries which have throwing methods, wrap these in try-catch. Will remove exceptions on REST side at a later point in time.
- Add a error macro header, eventually will move into shared header.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,9 @@ option (DO_INCLUDE_PLUGINS "Build subproject plugins" OFF)
 option (DO_INCLUDE_SDK "Build subproject sdk-cpp" OFF)
 
 option (DO_BUILD_TESTS "Set DO_BUILD_TESTS to OFF to skip building tests." ON)
+option (DO_ENABLE_EXCEPTIONS "Set DO_ENABLE_EXCEPTIONS to ON to disable compiling with exceptions" OFF)
+
+message(STATUS "Exceptions Disabled: ${DO_ENABLE_EXCEPTIONS}")
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
@@ -63,6 +66,10 @@ endif()
 
 # PIC (Position Independent Code) ensures .a files can be linked to executables that have PIE enabled
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+if (DO_ENABLE_EXCEPTIONS)
+    #TODO(jimson): There is no cmake option for disabling exceptions, so we must set a compiler-specific flag if we want to enforce disabling exceptions
+endif()
 
 # TODO: msvc and gcc support different compiler/linker flags, need to go through all flags and verify what's needed for each platform
 # Certain flags cause build issues because the flag itself isn't supported on msvc
--- a/sdk-cpp/include/do_download.h
+++ b/sdk-cpp/include/do_download.h
@@ -10,9 +10,7 @@
 #include <string>
 #include "do_download_status.h"
 
-#if defined(DO_INTERFACE_COM)
 #include "do_download_property.h"
-#endif
 
 namespace microsoft
 {
@@ -29,48 +27,42 @@ public:
     download(const std::string& uri, const std::string& downloadFilePath);
     ~download();
 
+#if (DO_ENABLE_EXCEPTIONS)
     void start();
     void pause();
     void resume();
     void finalize();
     void abort();
-
     download_status get_status() const;
 
     void start_and_wait_until_completion(std::chrono::seconds timeoutSecs = std::chrono::hours(24));
     void start_and_wait_until_completion(const std::atomic_bool& isCancelled, std::chrono::seconds timeoutSecs = std::chrono::hours(24));
-
     static void download_url_to_path(const std::string& uri, const std::string& downloadFilePath, std::chrono::seconds timeoutSecs = std::chrono::hours(24));
     static void download_url_to_path(const std::string& uri, const std::string& downloadFilePath, const std::atomic_bool& isCancelled, std::chrono::seconds timeoutSecs = std::chrono::hours(24));
 
-#if defined(DO_INTERFACE_COM)
-    /* Sets a download property for the current download object
-    This function attempts to set a download property using the caller-supplied value. This method can throw
-    ~~~~
+    /*
     For devices running windows before 20H1, dosvc exposed a now-deprecated com interface for setting certain download properties.
     After 20H1, these properties were added to newer com interface, which this SDK is using.
     Attempting to set a download property on a version of windows earlier than 20H1 will not set the property and throw an exception with error code msdo::errc::do_e_unknown_property_id
-    ~~~~
-    @param key, enum class defined in do_download_property.h that defines which property to set
-    @param value, the download_property_value to set
     */
     void set_property(download_property key, const download_property_value& value);
-
-    /* Same as set_property() but exceptions are handled internally and returned as an int32_t error code
-    ~~~~
-    @param key, enum class defined in do_download_property.h that defines which property to set
-    @param value, the download_property_value to set
-    @returns int32_t error code value, 0 on success, int32_t error code corresponding to the error otherwise.
-    */
-    int32_t set_property_nothrow(download_property key, const download_property_value& value) noexcept;
-
-    /* Gets the value for the specified download_property key for the current download object
-    ~~~~
-    @param key, the desired download_property to retrieve for this download object
-    @returns the download_property_value corresponding to the supplied key
-    */
-    download_property_value get_property(download_property key);
+    void get_property(download_property key, download_property_value& value);
 #endif
+
+    int32_t start_nothrow() noexcept;
+    int32_t pause_nothrow() noexcept;
+    int32_t resume_nothrow() noexcept;
+    int32_t finalize_nothrow() noexcept;
+    int32_t abort_nothrow() noexcept;
+    int32_t get_status_nothrow(download_status& status) noexcept;
+
+    int32_t start_and_wait_until_completion_nothrow(std::chrono::seconds timeoutSecs = std::chrono::hours(24)) noexcept;
+    int32_t start_and_wait_until_completion_nothrow(const std::atomic_bool& isCancelled, std::chrono::seconds timeoutSecs = std::chrono::hours(24)) noexcept;
+    static int32_t download_url_to_path_nothrow(const std::string& uri, const std::string& downloadFilePath, std::chrono::seconds timeoutSecs = std::chrono::hours(24)) noexcept;
+    static int32_t download_url_to_path_nothrow(const std::string& uri, const std::string& downloadFilePath, const std::atomic_bool& isCancelled, std::chrono::seconds timeoutSecs = std::chrono::hours(24)) noexcept;
+
+    int32_t set_property_nothrow(download_property key, const download_property_value& value) noexcept;
+    int32_t get_property_nothrow(download_property key, download_property_value& value) noexcept;
 
 private:
     std::shared_ptr<details::IDownload> _download;
@@ -78,4 +70,4 @@ private:
 
 } // namespace deliveryoptimization
 } // namespace microsoft
-#endif
+#endif // _DELIVERY_OPTIMIZATION_DO_DOWNLOAD_H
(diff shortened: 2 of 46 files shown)


CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ option (DO_INCLUDE_PLUGINS "Build subproject plugins" OFF)
 option (DO_INCLUDE_SDK "Build subproject sdk-cpp" OFF)
 
 option (DO_BUILD_TESTS "Set DO_BUILD_TESTS to OFF to skip building tests." ON)
+option (DO_ENABLE_EXCEPTIONS "Set DO_ENABLE_EXCEPTIONS to ON to disable compiling with exceptions" OFF)
```
shishirb-MSFT, line 18: >DO_ENABLE_EXCEPTIONS [](http://example.com/codeflow?start=8&length=20)

Since this applies to the sdk only, how about moving all changes in this file to the sdk cmakelists.txt?

shishirb-MSFT, line 18: >OFF [](http://example.com/codeflow?start=99&length=3)

I would default this to ON and let Edge disable it.

CMakeLists.txt
```diff
@@ -64,6 +67,10 @@ endif()
 # PIC (Position Independent Code) ensures .a files can be linked to executables that have PIE enabled
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+if (DO_ENABLE_EXCEPTIONS)
+    #TODO(jimson): There is no cmake option for disabling exceptions, so we must set a compiler-specific flag if we want to enforce disabling exceptions
```
shishirb-MSFT, line 71: >disabling [](http://example.com/codeflow?start=132&length=9)

Does the chronium build really disable exceptions in the compiler? IMO we should not worry about it since the builder can configure cmake over command line and set whatever compiler options they want.

sdk-cpp/include/do_download.h
```diff
@@ -45,9 +46,23 @@ class download
     Attempting to set a download property on a version of windows earlier than 20H1 will not set the property and throw an exception with error code msdo::errc::do_e_unknown_property_id
     */
     void set_property(download_property key, const download_property_value& value);
-    download_property_value get_property(download_property key);
+    void get_property(download_property key, download_property_value& value);
```
shishirb-MSFT, line 49: >void [](http://example.com/codeflow?start=4&length=4)

This should return the value instead, no?

sdk-cpp/include/do_error_macros.h
```diff
@@ -0,0 +1,54 @@
...
+    if(FAILED(__hr)) return __hr;   \
+}
+#endif
+
+#ifndef FACILITY_DELIVERY_OPTIMIZATION 
+#define FACILITY_DELIVERY_OPTIMIZATION   208
+#endif
+
+//TODO(jimson) Looks like these conversions are causing test issues with tests
+#define DO_ERROR_FROM_SYSTEM_ERROR(x) (int32_t)(0xC0000000 | (0xD0 << 16) | ((int32_t)(x) & 0x0000FFFF))
```
shishirb-MSFT, line 40: >0xD0 [](http://example.com/codeflow?start=62&length=4)

FACILITY_DELIVERY_OPTIMIZATION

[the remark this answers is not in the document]
shishirb-MSFT: [http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions)

sdk-cpp/include/do_exceptions.h
```diff
@@ -4,10 +4,13 @@
 #ifndef _DELIVERY_OPTIMIZATION_DO_EXCEPTIONS_H
```
shishirb-MSFT, line 4: >_DELIVERY_OPTIMIZATION_DO_EXCEPTIONS_H [](http://example.com/codeflow?start=8&length=38)

Considering renaming this file now that exceptions can be turned off, to not confuse users. do_errors.h?

sdk-cpp/src/do_download.cpp
```diff
@@ -84,78 +165,65 @@ void download::start_and_wait_until_completion(const std::atomic_bool& isCancell
...
+        RETURN_IF_FAILED(finalize_nothrow());
     }
     else
     {
-        abort();
+        RETURN_IF_FAILED(abort_nothrow());
         if (isCancelled)
         {
-            msdod::ThrowException(std::errc::operation_canceled);
+            return DO_ERROR_FROM_STD_ERROR(std::errc::operation_canceled);
```
shishirb-MSFT, line 182: >DO_ERROR_FROM_STD_ERROR [](http://example.com/codeflow?start=19&length=23)

When we return std::error_code instead, we don't have to perform these conversions. Tests will be happy too.

CMakeLists.txt
shishirb-MSFT: They probably don't need to do anything since they don't use cmake and by default DO_ENABLE_EXCEPTIONS will not be defined in their build system.

sdk-cpp/src/do_exceptions.cpp
```diff
@@ -1,5 +1,4 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
```
shishirb-MSFT, line 2: Retain copyright. You'll also need to scope the #if when implementing std::error_code return type.

sdk-cpp/src/internal/do_noncopyable.h
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#ifndef _DELIVERY_OPTIMIZATION_DO_NONCOPYABLE_H
+#ifndef _DELIVERY_OPTIMIZATION_DNONCOPYABLE_H
```
shishirb-MSFT, line 4: >_DELIVERY_OPTIMIZATION_DNONCOPYABLE_H [](http://example.com/codeflow?start=8&length=37)

revert

sdk-cpp/src/internal/com/do_download_property_internal.cpp
```diff
@@ -19,77 +19,85 @@ std::wstring UTF8toWstr(const char* str, size_t cch = 0)
...
+int32_t CDownloadPropertyValueInternal::Init(std::vector<unsigned char>& val) noexcept
 {
-    throw errc::e_not_impl;
+    return static_cast<int32_t>(errc::e_not_impl);
 };
 
-CDownloadPropertyValueInternal::CDownloadPropertyValueInternal(const download_property_value::status_callback_t& val)
+int32_t CDownloadPropertyValueInternal::Init(const download_property_value::status_callback_t& val) noexcept
 {
+    V_VT(&_var) = VT_EMPTY;
```
shishirb-MSFT, line 81: >V_VT(&_var) = VT_EMPTY; [](http://example.com/codeflow?start=4&length=23)

Why not use VariantInit? Standard/conventional way.

shishirb-MSFT: Since the std::vector<unsigned char> one also requires this, I suggest implementing the default constructor and calling VariantInit there. The constructor on the rest side can use = default in its .cpp file.

sdk-cpp/src/internal/com/do_download_property_internal.cpp
```diff
@@ -19,77 +19,85 @@ std::wstring UTF8toWstr(const char* str, size_t cch = 0)
...
 
 CDownloadPropertyValueInternal::CDownloadPropertyValueInternal(const CDownloadPropertyValueInternal& rhs)
 {
-    const auto res = VariantCopy(&_var, &rhs._var);
-#if (!DO_DISABLE_EXCEPTIONS)
-    microsoft::deliveryoptimization::throw_if_fail(res);
+#ifdef DEBUG
+    assert(SUCCEEDED(VariantCopy(&_var, &rhs._var)));
+#else:
+    (void)VariantCopy(&_var, &rhs._var);
```
shishirb-MSFT, line 100: >void [](http://example.com/codeflow?start=5&length=4)

Silently ignores errors. I suggest calling std::terminate instead when this fails (memory alloc failure).

sdk-cpp/src/internal/rest/download_impl.cpp
```diff
@@ -24,121 +24,145 @@ namespace deliveryoptimization
...
+                }
+                if (retryAttempts < g_maxNumRetryAttempts - 1)
+                {
+                    std::this_thread::sleep_for(1s);
+                }
             }
         }
+        return static_cast<int32_t>(msdo::errc::no_service);
+    }
+    catch (msdo::exception& e)
```
shishirb-MSFT, line 61: >msdo::exception [](http://example.com/codeflow?start=11&length=15)

All invoked methods here throw only this exception type?

sdk-cpp/src/do_exceptions.cpp
```diff
@@ -1,5 +1,4 @@
...
+#if (DO_ENABLE_EXCEPTIONS)
```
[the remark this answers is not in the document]
shishirb-MSFT: Copyright header is removed in some other files too. Intentional?

sdk-cpp/tests/com/download_properties_tests.cpp
```diff
@@ -77,22 +79,26 @@ TEST_F(DownloadPropertyTestsDOSVC, SmallDownloadWithPhfDigestandCvTest)
 
     std::vector<int32_t> expectedErrors = { 0, static_cast<int32_t>(msdo::errc::do_e_unknown_property_id) };
 
-    msdo::download_property_value integrityCheckMandatory(true);
+    std::unique_ptr<msdo::download_property_value> integrityCheckMandatory = std::make_unique<msdo::download_property_value>();
```
shishirb-MSFT, line 82: >make_unique [](http://example.com/codeflow?start=82&length=11)

Why allocate and then call make() again?

sdk-cpp/src/do_download.cpp
```diff
@@ -22,39 +21,44 @@ namespace deliveryoptimization
 
 download::download(const std::string& uri, const std::string& downloadFilePath)
 {
-    _download = std::make_shared<msdod::CDownloadImpl>(uri, downloadFilePath);
+    _download = std::make_shared<msdod::CDownloadImpl>();
+    _download->Init(uri, downloadFilePath);
```
shishirb-MSFT, line 25: >Init [](http://example.com/codeflow?start=15&length=4)

Ignores errors from the Init method. Do you intend to handle it in a separate PR?

sdk-cpp/include/do_download.h
```diff
@@ -45,9 +46,23 @@ class download
...
+#endif
+
+    int32_t start_nothrow() noexcept;
```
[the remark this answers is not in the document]
jimson-msft: I plan on adding a custom error class in the future PR and returning that custom error code, this was moreso to just get the code factored correctly.

sdk-cpp/include/do_error_macros.h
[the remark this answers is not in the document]
jimson-msft: Won't fix here, will reserve that for when we move these macros into the shared error macro header

sdk-cpp/include/do_download.h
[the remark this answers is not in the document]
shishirb-MSFT: Cool. Was gonna ask whether more PRs are coming.

sdk-cpp/include/do_download.h
jimson-msft: Yeah I'll have it as a return val instead of an out param here

CMakeLists.txt
shishirb-MSFT: I see this is copied to sdk-cpp/cmakelists.txt. Delete from here?

sdk-cpp/src/internal/do_noncopyable.h
shishirb-MSFT: fyi, not reverted yet

sdk-cpp/src/internal/com/do_download_property_internal.cpp
jimson-msft: Noted, fixing in next PR

sdk-cpp/src/internal/rest/download_impl.cpp
jimson-msft: Actually the uri builder invoked methods don't throw exceptions as far as I can tell.

PR #89, opened 2021-10-04
Templatize Github release jobs per platform & Tar the .deb output

jimson-msft:
- Tested this on a dry-run and release now contains 4 tar files respective to each platform
- It's a little messy the way this pipeline does the .tar file creation so that can be cleaned up in the future

PR #90, opened 2021-10-07
Templatize Github release jobs per platform & Tar the .deb output

shishirb-MSFT:
Cherry-picked from #89


shishirb-MSFT:
I think we can release v0.8.1 with the same internal component versions, with a short note saying this is an improvement in how artifacts are packaged. Any reason not to do this?

jimson-msft:
Why not just re-release 0.8.0? The release is already a little confusing with .deb copies

---
In reply to: [937386347](https://github.com/microsoft/do-client/pull/90#issuecomment-937386347) [](http://example.com/codeflow?ancestors=937386347,937386347)

PR #93, opened 2021-10-08
Cleanup comments

jimson-msft:


azure-pipelines/build/linux/du/templates/doclient-lite-native-steps.yml
```diff
@@ -82,16 +82,3 @@ steps:
     ArtifactName: 'deliveryoptimization-agent'
     publishLocation: 'Container'
 
-# TODO(jimson) Azure artifacts are no longer free to publish to, this task will fail as a result, bug to resolve issue here:
```
jimson-msft, line 85: Im removing these stubs, they're in the commit history if we need to bring them back. Also since you turned on subscription billing for our ADO project, we can theoretically use this feature now.

Nick/Oscar's team own testing going forward, so if they want to use ADO package feeds to retrieve artifacts, we can bring this back.

sdk-cpp/src/internal/com/deliveryoptimization/do.hpp
```diff
@@ -80,10 +80,10 @@ typedef enum _DODownloadProperty
     DODownloadProperty_IntegrityCheckInfo,
     DODownloadProperty_IntegrityCheckMandatory,
     DODownloadProperty_TotalSizeBytes,
-    DODownloadProperty_DisallowOnCellular,                  // TODO: Document this (Win11)
```
shishirb-MSFT, line 83: >DODownloadProperty_DisallowOnCellular [](http://example.com/codeflow?start=4&length=37)

It maybe useful to keep the min build required for these new properties, as comments. Win11/22000 for the first two and the last two requires pre-release builds prior to release of next Win11 version.

shishirb-MSFT (approved):
:shipit:

azure-pipelines/build/linux/du/templates/dopapt-native-steps.yml
```diff
@@ -72,17 +72,3 @@ steps:
...
-#  - task: UniversalPackages@0
-#    inputs:
-#      command: 'publish'
-#      publishDirectory: '$(Build.ArtifactStagingDirectory)'
-#      feedsToUsePublish: 'internal'
-#      vstsFeedPublish: 'a6e08e1f-d299-4d2f-aaa5-db7ddde849e0'
-#      publishPackageMetadata: false
-#      vstsFeedPackagePublish: 'deliveryoptimization-plugin-apt'
-#      versionOption: 'custom'
-#      versionPublish: '$(test.package.version)'
```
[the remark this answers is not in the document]
jimson-msft: Nope - they do not. Good point though

sdk-cpp/src/internal/com/deliveryoptimization/do.hpp
jimson-msft: I'll add a comment for the first two, I'm going to delete the last two for now. By the time these become available the windows sdk will have shipped with these two.

PR #94, opened 2021-10-12
Fix mac builds

jimson-msft:


sdk-cpp/src/internal/rest/do_config_internal.cpp
```diff
@@ -16,8 +14,13 @@
 #include <boost/property_tree/json_parser.hpp>
 #include "do_persistence.h"
 #include "do_version.h"
+#elif defined(DO_CLIENT_DOSVC)
```
shishirb-MSFT, line 17: >DO_CLIENT_DOSVC [](http://example.com/codeflow?start=14&length=15)

It is odd that only the dosvc build requires this header.

PR #95, opened 2021-10-13
Use std::error_code with custom error category

jimson-msft:


sdk-cpp/include/do_errors.h
```diff
@@ -29,22 +29,52 @@ enum class errc : int32_t
...
+    /*
+    const error_category& category() const noexcept
+    {
+        return _category;
+    }
+    */
+
+private:
+    int32_t _code;
+    //error_category _category;
```
shishirb-MSFT, line 71: >error_category [](http://example.com/codeflow?start=6&length=14)

IIRC, this should be a reference. Commonly, a reference to a static instance. See the constructor here: [https://en.cppreference.com/w/cpp/error/error_code/error_code](https://en.cppreference.com/w/cpp/error/error_code/error_code)

and cpprest's category here:
[https://github.com/microsoft/cpprestsdk/blob/master/Release/src/utilities/asyncrt_utils.cpp](https://github.com/microsoft/cpprestsdk/blob/master/Release/src/utilities/asyncrt_utils.cpp)

sdk-cpp/include/do_error_macros.h
```diff
@@ -4,49 +4,30 @@
...
-//TODO(jimson) Looks like these conversions are causing test issues with tests
+//TODO(jimson): Do we really need these after switching to error code class? The error_category should help us differentiate what type of error it is
 #define DO_ERROR_FROM_SYSTEM_ERROR(x) (int32_t)(0xC0000000 | (FACILITY_DELIVERY_OPTIMIZATION << 16) | ((int32_t)(x) & 0x0000FFFF))
 #define DO_ERROR_FROM_STD_ERROR(x) ((int32_t)(x) <= 0 ? ((int32_t)(x)) : ((int32_t) (((int32_t)(x) & 0x0000FFFF) | (FACILITY_DELIVERY_OPTIMIZATION << 16) | 0x80000000)))
 
-#ifndef FAILED
-#define FAILED(hr) (((int32_t)(hr)) < 0)
-#endif
+#define DO_RETURN_IF_FAILED(code)  {  \
+    if(DO_FAILED(code)) return code;   \
```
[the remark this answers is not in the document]
shishirb-MSFT: I mean the way it captures hr into __hr and uses __hr later.

sdk-cpp/include/do_error_macros.h
```diff
@@ -4,49 +4,30 @@
...
+}
 
-#ifndef SUCCEEDED
-#define SUCCEEDED(hr) (((int32_t)(hr)) >= 0)
-#endif
+#define DO_OK error_code(0)
+
+#define DO_SUCCEEDED(code) (((int32_t)(code.value())) < 0)
```
shishirb-MSFT, line 29: >((int32_t)(code.value())) < 0 [](http://example.com/codeflow?start=28&length=29)

Since we expect only error_code type here, we should use its operator bool instead. At which point, I question the utility of these two macros.

shishirb-MSFT: Read this comment after my other one in do_errors.h about using std::error_code.

sdk-cpp/include/do_error_macros.h
```diff
@@ -4,49 +4,30 @@
...
-    if(FAILED(__hr)) return __hr;   \
-}
-#endif
-
 #ifndef FACILITY_DELIVERY_OPTIMIZATION 
 #define FACILITY_DELIVERY_OPTIMIZATION   208
 #endif
 
-//TODO(jimson) Looks like these conversions are causing test issues with tests
+//TODO(jimson): Do we really need these after switching to error code class? The error_category should help us differentiate what type of error it is
```
shishirb-MSFT, line 19: >need [](http://example.com/codeflow?start=29&length=4)

Agree we don't need these anymore.

shishirb-MSFT (approved):
:shipit:

sdk-cpp/include/do_error_macros.h
jimson-msft: ah nvm, thinking about your comment in do_errors.h regarding using error_category

jimson-msft:
> > ```
> > int32_t error_code() const;
> > ```
> 
> IMO this is redundant since caller can get it from std::error_code. We should remove this and rename get_error_code() to error_code().
> 
> In reply to: [942778026](https://github.com/microsoft/do-client/pull/95#issuecomment-942778026)
> 
> In reply to: [942778026](https://github.com/microsoft/do-client/pull/95#issuecomment-942778026)
> 
> In reply to: [942778026](https://github.com/microsoft/do-client/pull/95#issuecomment-942778026)
> 
> In reply to: [942778026](https://github.com/microsoft/do-client/pull/95#issuecomment-942778026)
> 
> Refers to: sdk-cpp/include/do_errors.h:84 in [efea6d9](https://github.com/microsoft/do-client/commit/efea6d9ba08fe7ee387b2844d65a3b30abba7b6d). [](commit_id = [efea6d9](https://github.com/microsoft/do-client/commit/efea6d9ba08fe7ee387b2844d65a3b30abba7b6d), deletion_comment = False)

Theres s

sdk-cpp/include/do_error_macros.h
jimson-msft: Let's still apply the facility code where needed for DU purposes, and make use of error_category

sdk-cpp/include/do_errors.h
jimson-msft: sounds good to me

sdk-cpp/src/do_errors.cpp
```diff
@@ -1,26 +1,38 @@
...
-const error_category& error_category_instance()
+const error_category& do_category()
 {
     static error_category instance;
     return instance;
 }
 
+std::error_code make_error_code(std::errc e)
+{
+    return std::error_code(DO_ERROR_FROM_STD_ERROR(e), do_category());
```
shishirb-MSFT, line 19: >DO_ERROR_FROM_STD_ERROR [](http://example.com/codeflow?start=27&length=23)

Need to fix tests to avoid doing this but it's okay for now.

shishirb-MSFT: I'll sign off.

shishirb-MSFT (approved):
:shipit:

sdk-cpp/src/do_errors.cpp
jimson-msft: Hmm I'm not sure I follow - I think we do need to do this in order to make use of the same error code macros/convention. the std::errc enum are a set of positive integer values, and so need to call DO_ERROR_FROM_STD_ERROR to make them hresult-like.

PR #96, opened 2021-10-20
Support Ubuntu 20.04 on amd64 and arm64

jimson-msft:


shishirb-MSFT:
>     if [[ $"PLATFORM" != "unknown" ]]

$ should go after the opening quote "

---
Refers to: build/scripts/bootstrap.sh:67 in d7a831e. [](commit_id = d7a831ec674d3e1fadf8fffb3c8445079aac5d40, deletion_comment = False)

build/docker/ubuntu2004/amd64/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Dockerfile for building DO apt plugin for linux-arm (64bit).
```
shishirb-MSFT, line 1: Comment needs updating. In other Dockerfiles as well. Not urgent..

build/docker/ubuntu2004/amd64/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# First, install the docker extension for VSCode. Then you can right-click on this file
+# and choose Build Image. Give it a name and it will build the image.
+#
+# Open interactive terminal into the image in a container:
+# docker run -ti --rm --entrypoint=/bin/bash -v <project root dir>:/code -v <build root dir>:/build <image_name>
+# Example:
+# docker run -ti --rm --entrypoint=/bin/bash -v D:\do-client-lite:/code -v D:\temp\build_client_lite\arm-linux-debug:/build custom-ubuntu2004-arm64
+
+FROM amd64/ubuntu:20.04
```
shishirb-MSFT, line 10: >amd64 [](http://example.com/codeflow?start=5&length=5)

We recently came to the conclusion that we'll go native where possible. Isn't a native 20.04 image available in 1ES?

jimson-msft: Yeah I had considered that in an earlier iteration. We can do a native build, but this requires us to setup another 1ES image + another set of pipeline yml files specifically for ubuntu 20.04.

The current 1ES image is named:
do-adu-build-<Semantic Version Number of current (really meaning subsequent) release>

So now we'll have two 1ES images for example, per version:
do-adu-build-ubuntu1804-0.8.0
do-adu-build-ubuntu2004-0.8.0

I don't think it's worth maintaining duplicate vm images for native builds, if we trust that the docker builds are doing what they're supposed to.
Another question may be - if we do native builds for Ubuntu 20.04, do we also run tests there? IMO it's unnecessary since we rely on Ubuntu 1804 for tests.

IMO we should rely on still having a single build image to support all our builds, and only run tests natively, and have CI/CD triggers for this native build to save on compute resources.

PR #97, opened 2021-10-21
Should copy .dbg symbols for debug builds not minsizerel

jimson-msft:


shishirb-MSFT:
>       client-lite/deliveryoptimization-agent.dbg

In non-minsizerel builds, the debug symbols get baked into the binary itself.

---
In reply to: [948868102](https://github.com/microsoft/do-client/pull/97#issuecomment-948868102) [](http://example.com/codeflow?ancestors=948868102)

---
Refers to: azure-pipelines/build/linux/du/templates/doclient-lite-docker-steps.yml:46 in c214d12. [](commit_id = c214d1267a0036f029109525ef21550895c0d4e2, deletion_comment = False)

PR #103, opened 2021-11-30
Implement Generator parser argument

mdavis777:
Currently the generator parser argument causes the build to error out as a setter was never implemented.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -154,6 +154,7 @@ class BuildRunnerBase(object):
         self.cmake_target = None
         self.project = None
         self.build_directory = None
+        self._generator = None
 
         if (script_args.cmaketarget is None):
             self.cmake_target = "all"
@@ -227,8 +228,8 @@ class BuildRunnerBase(object):
             None if default generator should be used (not recommended).
         """
         if is_running_on_linux():
-            return "Ninja"
-        return None
+            self._generator = "Ninja"
+        return self._generator
 
     @property
     def platform(self):
@@ -434,7 +435,11 @@ class LinuxBuildRunner(BuildRunnerBase):
 
     @property
     def generator(self):
-        return 'Ninja'
+        return self._generator or 'Ninja'
+
+    @generator.setter
+    def generator(self, value):
+        self._generator = value
 
     @property
     def generate_options(self):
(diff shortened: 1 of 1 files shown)


build/build.py
```diff
@@ -434,7 +435,11 @@ def platform(self):
 
     @property
     def generator(self):
-        return 'Ninja'
+        return self._generator or 'Ninja'
+
+    @generator.setter
+    def generator(self, value):
```
shishirb-MSFT, line 441: >generator [](http://example.com/codeflow?start=8&length=9)

Instead of repeating the setter in derived classes, can you move it to the base class BuildRunnerBase?

shishirb-MSFT:
>             self.generator = self.script_args.generator

Looks like the generator command line arg is already being consumed here. The issue maybe that self.generator is not returned from the getter methods.

Can you rework the fix to use self.generator instead of adding a new self._generator member?

---
Refers to: build/build.py:181 in fd0f527. [](commit_id = fd0f5274552034c28c6bc765e6bd6914fcf6bf12, deletion_comment = False)

mdavis777:
> > ```
> >         self.generator = self.script_args.generator
> > ```
> 
> Looks like the generator command line arg is already being consumed here. The issue maybe that self.generator is not returned from the getter methods.
> 
> Can you rework the fix to use self.generator instead of adding a new self._generator member?
> 
> Refers to: build/build.py:181 in [fd0f527](https://github.com/microsoft/do-client/commit/fd0f5274552034c28c6bc765e6bd6914fcf6bf12). [](commit_id = [fd0f527](https://github.com/microsoft/do-client/commit/fd0f5274552034c28c6bc765e6bd6914fcf6bf12), deletion_comment = False)

No because self.generator is the function being called as the getter.  If I change it to self.generator it is an infinite loop

build/build.py
mdavis777: I could but then I would still have to have the derived call settings there to call the parents setter which results in more code.  
`    @generator.setter
    def generator(self, value):
        super(LinuxBuildRunner, type(self)).generator.fset(self, value)`

If that is what you want just let me know and I will make the changes

shishirb-MSFT:
Please see this commit: [https://github.com/microsoft/do-client/commit/d714befc414e0967fcae65e855fb9920ea99b8d3](https://github.com/microsoft/do-client/commit/d714befc414e0967fcae65e855fb9920ea99b8d3)

Base class provides the user specified cmdline argument and derived classes provide the platform-specific default when needed. This was the intended design but wasn't implemented correctly given that the Windows and Mac variations were added later.

---
In reply to: [985878609](https://github.com/microsoft/do-client/pull/103#issuecomment-985878609) [](http://example.com/codeflow?ancestors=985878609,985878609)

shishirb-MSFT:
@mdavis777 This has been fixed in main branch (see my other comment about the alternate fix). I will close this PR but please let me know if something is still missing for your use case.

PR #112, opened 2022-02-11
[DO]: Statically initialize downloader object for no-exception case

neagra1:


sdk-cpp/include/do_download.h
```diff
@@ -50,7 +50,7 @@ class download
     download_property_value get_property(download_property key);
 #endif
 
-    static std::error_code make_nothrow(const std::string& uri, const std::string& downloadFilePath, download& out) noexcept;
+    static download make_nothrow(const std::string& uri, const std::string& downloadFilePath, std::error_code& error) noexcept;
```
shishirb-MSFT, line 53: >error [](http://example.com/codeflow?start=111&length=5)

This makes our API inconsistent because all other methods use return value for the error code instead of an out param.

One option I was thinking of was to use lazy-init, i.e., call Init() from the other existing public methods that can return an error code (no need for make/make_nothrow methods with this option).

Let me know if you want me to do this.

neagra1: Yes, but I don't see an issue in it being different. Other way was to just make the empty constructor public, but I remember you mentioned, you don't want that. I don;t undersatnd how lazy-init can solve the issue we have of not having any way to initialize a downloader object in no-throw part.

Can you give me a function declaration of what you are thinking?  And what I can use in Edge to create a unique downloader object for each download instance ?

neagra1: Also, I don't think having init from any other public method in this file is going to solve our purpose as we need to set some downloader properties before starting the download, so there needs to be a method to return a downloader object.

shishirb-MSFT:
#113 supersedes this PR.

PR #124, opened 2022-08-04
Adjust MCC config and http timeouts

shishirb-MSFT:


shishirb-MSFT:
Timeouts have been adjusted as a mitigation. Hence, leaving the MCC auto-configuration via IoT conn string as-is.

PR #130, opened 2022-08-17
SDK, Plugin: Remove exceptions from public API surface

shishirb-MSFT:


plugins/linux-apt/main.cpp
```diff
@@ -194,7 +193,11 @@ DownloadResult SimpleDownload(const std::string& url, const std::string& destFil
         tempPath = GenerateUniqueCachePath();
         LogDebug("Using temp file %s for %s", tempPath.data(), destFilePath.data());
 
-        msdo::download::download_url_to_path(url, tempPath);
+        std::error_code errorCode = msdo::download::download_url_to_path(url, tempPath);
+        if (errorCode)
+        {
+            throw DOPluginException("Download hit an error: %d, %s", errorCode.value(), errorCode.message().c_str());
```
JeffMill, line 199: nit: string messages add bloat to the binary. Keep terse but understandable, e.g.

`"Download hit an error: %d, %s"` -> `"download_url_to_path error: %d, %s"`

JeffMill: similarly on 215.

sdk-cpp/include/do_download_property.h
```diff
@@ -65,43 +65,27 @@ CDownloadImpl is declared as a friend class because it needs to access the platf
...
-    static std::error_code make_nothrow(std::vector<unsigned char>& val, download_property_value& out);
-    static std::error_code make_nothrow(const status_callback_t& val, download_property_value& out);
-
-    std::error_code as_nothrow(bool& val) const noexcept;
-    std::error_code as_nothrow(uint32_t& val) const noexcept;
-    std::error_code as_nothrow(uint64_t& val) const noexcept;
-    std::error_code as_nothrow(std::string& val) const noexcept;
-    std::error_code as_nothrow(std::vector<unsigned char>& val) const noexcept;
-    std::error_code as_nothrow(status_callback_t& val) const noexcept;
+    static std::error_code make(const std::string& val, download_property_value& out);
```
JeffMill, line 76: for all of these classes you're changing methods to noexcept, should the classes themselves ensure they dont throw, or are you just concerned about the methods -- for example, should these classes delete the copy ctor, move operator, assignment operator, etc to avoid exceptions being thrown from them?

Alternately (and a bigger task) is that if these classes are just method implementations (e.g. all static methods), then a namespace with methods in it might be more suitable.

sdk-cpp/src/internal/com/do_download_property_internal.cpp
```diff
@@ -109,11 +109,7 @@ CDownloadPropertyValueInternal::CDownloadPropertyValueInternal(const CDownloadPr
 #endif
```
JeffMill, line 109: related to my question earlier -- is the change now to not throw exceptions at all, or just to not surface exceptions?  I'm guessing it's the latter, but the previous implementation of this method was:

* Exceptions enabled: throw bad_alloc
* Exceptions disabled: terminate process

Now the behavior is (implied) Exceptions disabled: throw bad_alloc

JeffMill, line 109: related to my question earlier -- is the change now to not throw exceptions at all, or just to not surface exceptions?  I'm guessing it's the latter, but the previous implementation of this method was:

* Exceptions enabled: throw bad_alloc
* Exceptions disabled: terminate process

Now the behavior is (implied) Exceptions disabled: throw bad_alloc

sdk-cpp/tests/download_properties_tests.cpp
```diff
@@ -51,23 +42,35 @@ class DownloadPropertyTestsDOSVC_NoThrow : public ::testing::Test
         auto ec = op();
         VerifyError(ec.value(), expectedErrors);
     }
+#endif
 };
 
+// Only the DoSvc client (windows) supports get/set property at present
```
JeffMill, line 48: what's TimeOperation have to do with get/set property?

sdk-cpp/tests/test_data.cpp
```diff
@@ -16,6 +16,7 @@ const std::string g_tmpFileName = (boost::filesystem::temp_directory_path() / bo
 const std::string g_tmpFileName2 = (boost::filesystem::temp_directory_path() / boost::filesystem::path("docsdk_testfile2.txt")).string();
 const std::string g_tmpFileName3 = (boost::filesystem::temp_directory_path() / boost::filesystem::path("docsdk_testfile3.txt")).string();
 const std::string g_smallFileUrl = "http://main.oremdl.microsoft.com.nsatc.net/dotc/49c591d405d307e25e72a19f7e79b53d69f19954/43A54FC03C6A979E9AAEAE2493757D1429A5C8A8D093FB7B8103E8CC8DF7B6B6";
+const std::string g_smallFilePhfInfoJson = "{\"PiecesHashFileUrl\":\"https://eus2intdo101.blob.core.windows.net/partnercatalogcontainer/pc00051365650774478716C3536617271346B6B33563946436D6C794B6A516B2F74376751506F7A4933337472593D0000B9CCB33-51FB-4062-AFAC-08C8A088CB35.json\",\"HashOfHashes\":\"Q6VPwDxql56arq4kk3V9FCmlyKjQk/t7gQPozI33trY=\"}";
```
JeffMill, line 19: nit: R string would make this easier to read.

sdk-cpp/tests/test_helpers.h
```diff
@@ -8,21 +8,146 @@
...
+}
+}
+}
+
 class TestHelpers
 {
 public:
-    static void CleanTestDir();
-    static void CleanupWorkingDir();
+    static void CleanTestDir()
```
JeffMill, line 132: you're using C++17, right?  if so, boost isn't needed.

`std::filesystem::exists`, `std::filesystem::remove`

JeffMill (approved):

sdk-cpp/tests/download_properties_tests.cpp
shishirb-MSFT: Comment is meant for the ifdef. See the last test case here - TimeOperation is used to (indirectly) verify that the foreground/background property has taken effect by comparing the time taken to download.

sdk-cpp/tests/test_helpers.h
shishirb-MSFT: Still on C++14.

sdk-cpp/tests/test_data.cpp
shishirb-MSFT: Sweet, didn't know about it.

sdk-cpp/include/do_download_property.h
shishirb-MSFT: Only concerned about the methods here.

The static methods do need to call a method on a private member so I'll leave them within the class.

sdk-cpp/src/internal/com/do_download_property_internal.cpp
shishirb-MSFT: Just to not surface expected errors as exceptions. I initially retained std::terminate here but decided that is too heavy handed.

JeffMill (approved):

PR #135, opened 2022-09-08
fix/allow admin to override mcc config with empty string

andretoyama-msft:
BUG 40525399:

ADU DO Client: Allow admins to override MCC config with 'empty string' to disable MCC usage

### DIFF
--- a/client-lite/src/config/mcc_manager.cpp
+++ b/client-lite/src/config/mcc_manager.cpp
@@ -50,14 +50,18 @@ boost::optional<std::chrono::seconds> MCCManager::FallbackDelay()
 
 std::string MCCManager::GetHost()
 {
-    std::string mccHostName =_configManager.CacheHostServer();
-    if (mccHostName.empty())
+    boost::optional<std::string> mccHostNameOpt =_configManager.CacheHostServer();
+    std::string mccHostName;
+
+    if (!mccHostNameOpt.is_initialized())
     {
         const std::string connString = _configManager.IoTConnectionString();
         if (!connString.empty())
         {
             mccHostName = GetHostNameFromIoTConnectionString(connString.data());
         }
+    } else {
+        mccHostName = mccHostNameOpt.get();
     }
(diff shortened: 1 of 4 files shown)


client-lite/src/config/mcc_manager.cpp
```diff
@@ -50,14 +50,18 @@ boost::optional<std::chrono::seconds> MCCManager::FallbackDelay()
...
+    std::string mccHostName;
+
+    if (!mccHostNameOpt.is_initialized())
     {
         const std::string connString = _configManager.IoTConnectionString();
         if (!connString.empty())
         {
             mccHostName = GetHostNameFromIoTConnectionString(connString.data());
         }
+    } else {
```
shishirb-MSFT, line 63: >} else { [](http://example.com/codeflow?start=4&length=8)

else keyword must be placed on its own line:
}
else
{

client-lite/src/config/mcc_manager.cpp
```diff
@@ -50,14 +50,18 @@ boost::optional<std::chrono::seconds> MCCManager::FallbackDelay()
 
 std::string MCCManager::GetHost()
 {
-    std::string mccHostName =_configManager.CacheHostServer();
-    if (mccHostName.empty())
+    boost::optional<std::string> mccHostNameOpt =_configManager.CacheHostServer();
+    std::string mccHostName;
+
+    if (!mccHostNameOpt.is_initialized())
```
shishirb-MSFT, line 56: >!mccHostNameOpt.is_initialized() [](http://example.com/codeflow?start=8&length=32)

Flip the conditional to avoid the negation (improves readability):
if (mccHostNameOpt.is_initialized())
{
    mccHostName = mccHostNameOpt.get();
}
else
{
    ...
}

shishirb-MSFT:
Waiting for a unit test to be written.

shishirb-MSFT (approved):

Issue #62, opened 2021-08-26
DO-Agent uses to mutch RAM


Voxel07:
When the deployment is restarted after a field installation, the agent sometimes can't download the file and runs out of memory.

This happened before, when the DU-Agent was restarted during the download phase. This was fixed by removing old sandboxes and restarting do-service before starting the new download. The download can now be aborted and restarted as many times as wanted. But if DU-Agent isn't restarted after a failed Download, something goes wrong with DO-Agent

The Problem is, that this isn´t happening all the time. Sometimes the download succeeds many times in a row and then crashes again randomly. If I run top alongside, then there is always memory left.

# DU-LOG
```
root@fsimx8mm:~# su -m adu -c "AducIotAgent -l 0"
2021-08-26T12:53:33.1530Z [I] Agent (linux; 0.7.0-public-preview) starting. [main]
2021-08-26T12:53:33.1531Z [I] Agent built with handlers: fus/fsupdate. [main]
2021-08-26T12:53:33.1550Z [I] Attempting to create connection to IotHub using type: ADUC_ConnType_Device  [ADUC_DeviceClient_Create]
2021-08-26T12:53:33.1551Z [I] Initalizing PnP components. [ADUC_PnP_Components_Create]
2021-08-26T12:53:33.1552Z [I] ADUC agent started. Using IoT Hub Client SDK 1.6.0 [AzureDeviceUpdateCoreInterface_Create]
2021-08-26T12:53:33.1552Z [I] Calling ADUC_Register [ADUC_MethodCall_Register]
2021-08-26T12:53:33.1552Z [I] IoTHub Device Twin callback registered. [ADUC_DeviceClient_Create]
2021-08-26T12:53:33.1562Z [I] Agent running. [main]
2021-08-26T12:53:33.8327Z [D] IotHub connection status: 0, reason:6 [ADUC_ConnectionStatus_Callback]
2021-08-26T12:53:34.2350Z [D] ComponentName:azureDeviceUpdateAgent, propertyName:service [ADUC_PnP_ComponentClient_PropertyUpdate_Callback]
[...]

PR #136, opened 2022-09-09
Agent: Handle failures from callbacks during download

shishirb-MSFT:


shishirb-MSFT:
Fixes problem identified in #62

PR #140, opened 2022-09-26
Agent: Fix debian package upgrade issue

shishirb-MSFT:
* During an upgrade, postrm script deletes the 'do' group causing 'adu' membership to be removed.
* postrm script is now updated to perform cleanup only on remove and purge actions. postinst script already handled the case of existing user/group entities.
* Also move systemctl reset-failed to postinst as the more appropriate location. If retaining in prerm, we will have to take a call on whether to call it only on upgrade/abort-upgrade/etc.

### DIFF
--- a/client-lite/build/postrm.in.sh
+++ b/client-lite/build/postrm.in.sh
@@ -58,8 +58,7 @@ case "$1" in
     # upgrade: Removing user/group/directories during an upgrade is a mistake that can impact group membership, historical logs, etc.
     # abort-install, abort-upgrade: This project does not use any package install path that requires handling these cases.
     upgrade|abort-install|abort-upgrade)
-        exit 0
     ;;
 esac
 
-echo "Done!"
+echo "**** Done! ****"
--- a/client-lite/build/postrm.in.sh
+++ b/client-lite/build/postrm.in.sh
@@ -11,7 +11,7 @@ run_path=@docs_svc_run_dir_path@
 svc_name=@docs_svc_name@
 svc_config_path=@docs_systemd_cfg_path@
 
-echo "Running post-removal script for $svc_name, args: $1, $2"
+echo "**** Running post-removal script for $svc_name, args: $1 $2 ****"
 
 do_remove_daemon() {
     echo "Removing systemd unit file: $svc_config_path"
--- a/client-lite/build/postrm.in.sh
+++ b/client-lite/build/postrm.in.sh
@@ -11,29 +11,55 @@ run_path=@docs_svc_run_dir_path@
 svc_name=@docs_svc_name@
 svc_config_path=@docs_systemd_cfg_path@
 
-echo "Running post-removal script for $svc_name"
+echo "Running post-removal script for $svc_name, args: $1, $2"
 
-systemctl reset-failed $svc_name # Remove ourselves from the failed services list. No-op if never failed earlier.
+do_remove_daemon() {
+    echo "Removing systemd unit file: $svc_config_path"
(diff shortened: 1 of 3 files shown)


client-lite/build/postrm.in.sh
```diff
@@ -11,29 +11,55 @@ run_path=@docs_svc_run_dir_path@
 svc_name=@docs_svc_name@
 svc_config_path=@docs_systemd_cfg_path@
 
-echo "Running post-removal script for $svc_name"
+echo "**** Running post-removal script for $svc_name, args: $1 $2 ****"
 
-systemctl reset-failed $svc_name # Remove ourselves from the failed services list. No-op if never failed earlier.
+do_remove_daemon() {
+    echo "Removing systemd unit file: $svc_config_path"
+    rm $svc_config_path
```
jw-msft, line 18: There is no `set -e` in this file but it exists in postrm.in.sh.  Could add that, or avoid pitfalls with that and do explicit error checking via adding `|| return` on the end in function definitions or `|| exit` elsewhere (e.g. `mycmd myargs || return`), which will forward along the non-zero ret code when it fails.

I would encourage running shellcheck on these scripts.

client-lite/build/postrm.in.sh
```diff
@@ -11,29 +11,55 @@ run_path=@docs_svc_run_dir_path@
...
-echo "Removing run directory: $run_path"
-rm -rf $run_path
+    echo "Removing group: ${do_group_name}"
+    groupdel ${do_group_name}
+}
 
-echo "Removing user: ${do_user_name}"
-userdel ${do_user_name}
+do_remove_tasks() {
+    do_remove_daemon
```
jw-msft, line 41: `|| return` for these

client-lite/build/postrm.in.sh
```diff
@@ -11,29 +11,55 @@ run_path=@docs_svc_run_dir_path@
...
+        do_remove_tasks
+
+        echo "Removing config directory: $config_path"
+        rm -rf $config_path
+    ;;
+
+    # upgrade: Removing user/group/directories during an upgrade is a mistake that can impact group membership, historical logs, etc.
+    # abort-install, abort-upgrade: This project does not use any package install path that requires handling these cases.
+    upgrade|abort-install|abort-upgrade)
+        exit 0
```
jw-msft, line 61: nit: could be noop so it could echo "Done!" and exit with 0 implicitly.

client-lite/build/postrm.in.sh
shishirb-MSFT: I noticed the missing set -e but decided it is out of scope for this PR. I'll make such improvements in a separate PR to avoid introducing more bugs before GA.

client-lite/build/postrm.in.sh
shishirb-MSFT: Since there is no set -e in this yet, all steps are considered as best-effort.

client-lite/build/postrm.in.sh
shishirb-MSFT: +1 for shellcheck, thanks

jw-msft:
unless I'm missing something, it seems like postrm.in.sh and prerm.in.sh need `set -e` so it will return failure ret code when individual commands fail so that it does not leave the install/upgrade in a partial state.

shishirb-MSFT:
Agree about set -e but it is not in scope for this PR. Adding set -e requires more changes like deleting file/directory only if it exists.

PR #142, opened 2022-10-28
Updates for GA

shishirb-MSFT:


cmknox:
@shishirb-MSFT what will the version of the 'deliveryoptimization-plugin-apt' be?

shishirb-MSFT:
> @shishirb-MSFT what will the version of the 'deliveryoptimization-plugin-apt' be?

It will stay what it is now - 0.5.1

PR #149, opened 2022-12-27
Initial stage of DoClient Snap/Agent app and ubuntu core only flag

andretoyama-msft:
Initial commit for the [DELIVERABLE 41450767](https://microsoft.visualstudio.com/OS/_workitems/edit/41450767)

Features:
 - Initial Snapcraft.yaml file for Do Client snap build
 - Delivery Optimization Agent app built on the Do client snap
 - "--ubuntu-core-snap-only" flag to only build components and features required for the Ubuntu Core snap
 - Initial state of Network bind slots for communication with ADU snap


client-lite/src/exe/docs.cpp
```diff
@@ -139,8 +139,11 @@ HRESULT Run() try
 
     RestPortAdvertiser portAdvertiser(controller.Port());
     DoLogInfo("Port number written to %s", portAdvertiser.OutFilePath().data());
+    
+    #ifndef UBUNTU_CORE_SNAP_ONLY
```
shishirb-MSFT, line 143: Directives are always placed at the left margin, so reduce indentation by one level for all 3 lines here.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+    plugin: nil
+    source: .
+    override-build: |
+      ./build/scripts/bootstrap.sh --platform ubuntu2004 --install build
+
+  agent:
+    plugin: python
+    source: .
+    override-build: |
+      python3 ./build/build.py --project agent --ubuntu-core-snap-only
```
shishirb-MSFT, line 33: This can come later but something to keep in mind before calling this complete: figure out how to build debug and minsizerel versions for snap.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+confinement: devmode # use 'strict' once you have the right plugs and slots
+
+#####
+#
+# Keywords
+#
+# after  - Specifies part's dependencies.
+#              See https://snapcraft.io/docs/parts-lifecycle#heading--step-dependencies
+#                  https://snapcraft.io/docs/parts-lifecycle
+# plugin - Specifies a plugin required for building the part.
```
shishirb-MSFT, line 18: Would be useful to include the explanation for plugs and slots also.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+    restart-delay: 10s
+    plugs:
+      - network
+      - network-bind
+
+slots:
+  port-number:
+    interface: content
+    content: port-number
+    read: [ $SNAP_DATA/var/run/deliveryoptimization-agent/restport ]
```
shishirb-MSFT, line 76: The filename changes with every launch of the daemon, so can we mention read for the folder?
$SNAP_DATA/var/run/deliveryoptimization-agent/

snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+  config-file:
+    interface: content
+    content: config-file
+    read: [ $SNAP_DATA/etc/deliveryoptimization-agent/sdk-config.json ]
```
shishirb-MSFT, line 81: DU agent needs write permission for this config file.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+    command: bin/deliveryoptimization-agent
+    daemon: simple
+    refresh-mode: restart
+    restart-condition: always
+    restart-delay: 10s
+    plugs:
+      - network
+      - network-bind
+
+slots:
```
shishirb-MSFT, line 72: DU agent probably needs read permission for our log folder also. We'll discuss next week.

shishirb-MSFT:
The call to SetDOPathPermissions in proc_launch_helper.h also should be skipped if building for snap.

shishirb-MSFT (changes requested):

CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ option (DO_INCLUDE_PLUGINS "Build subproject plugins" OFF)
 option (DO_INCLUDE_SDK "Build subproject sdk-cpp" OFF)
 
 option (DO_BUILD_TESTS "Set DO_BUILD_TESTS to OFF to skip building tests." ON)
+option (DO_BUILD_FOR_SNAP "Enable DO Snap build option" ON)
```
shishirb-MSFT, line 18: The default should be OFF otherwise all builds will start generating snaps.

andretoyama-msft: oh ok my bad, I misunderstood the meaning of that

shishirb-MSFT (approved):

PR #152, opened 2023-01-26
Allow overriding cmake search path

andretoyama-msft:
[TASK 42848937](https://microsoft.visualstudio.com/OS/_workitems/edit/42848937)

Why is this change being made?
To add Support for SDK build and installation in the Delivery Optimization Snap

What is being changed?
Snapcraft yaml file to properly build and install the sdk in the correct folders

How was the change tested?
Checking if the libdeliveryoptimization file was in the install folder

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,9 @@ if (DO_PLATFORM_LINUX)
     # (Enable --trace-expand in cmake and look at lines from boost_headers-config.cmake and its callers.)
     # Workaround: force cmake to look in "/usr" since the /lib and /bin paths are either
     # not relevant for our usage or they are symlinks to /usr/lib and /usr/bin.
-    # set(CMAKE_PREFIX_PATH "/usr")
+    if (NOT CMAKE_PREFIX_PATH)
+        set(CMAKE_PREFIX_PATH "/usr")
+    endif () 
 endif (DO_PLATFORM_LINUX)
 
 if (DO_PLATFORM_WINDOWS AND DO_INCLUDE_SDK)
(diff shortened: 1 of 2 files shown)


build/build.py
```diff
@@ -98,6 +98,10 @@ def __init__(self):
             '--build-for-snap', dest='build_for_snap', action='store_true',
             help='Only build components and features those required for the Ubuntu Core snap'
         )
+        self.parser.add_argument(
+            '--install-prefix', dest='install_prefix', type=str,
```
shishirb-MSFT, line 102: Rename the option to --search-prefix. Install prefix is the next change to come.

build/build.py
```diff
@@ -444,6 +449,11 @@ def generate_options(self):
         if self.build_for_snap:
             generate_options.extend(["-DDO_BUILD_FOR_SNAP=1"])
 
+        if self.install_prefix:
+            generate_options.extend(["-DCMAKE_PREFIX_PATH={}".format(self.install_prefix)])
+        else:
+            generate_options.extend(["-DCMAKE_PREFIX_PATH=/usr"])
```
shishirb-MSFT, line 455: Let us handle the default in CMakeLists.txt itself. DU agent does not use build.py, so it will help keep the behavior same for them.

CMakeLists.txt
```diff
@@ -83,7 +83,7 @@ if (DO_PLATFORM_LINUX)
     # (Enable --trace-expand in cmake and look at lines from boost_headers-config.cmake and its callers.)
     # Workaround: force cmake to look in "/usr" since the /lib and /bin paths are either
     # not relevant for our usage or they are symlinks to /usr/lib and /usr/bin.
-    set(CMAKE_PREFIX_PATH "/usr")
+    # set(CMAKE_PREFIX_PATH "/usr")
```
shishirb-MSFT, line 86: As I mentioned in the other comment, let us set CMAKE_PREFIX_PATH only if it is empty here. Once changed, make sure you verify that overriding from build.py and using the default works properly.

shishirb-MSFT (approved):

PR #153, opened 2023-02-02
SDK: Add get_property support, remove callbacks and vectors from properties

JeffreySaathoff:


JeffreySaathoff:
@microsoft-github-policy-service agree company="Microsoft"

sdk-cpp/include/do_download_property.h
```diff
@@ -23,69 +22,84 @@ class CDownloadPropertyValueInternal;
...
 friend class details::CDownloadImpl;
 
 public:
-    using status_callback_t = std::function<void(download&, download_status&)>;
-
     download_property_value();
     ~download_property_value() = default;
 
     static std::error_code make(const std::string& val, download_property_value& out);
+    static std::error_code make(const char* val, download_property_value& out) { return make(std::string(val), out); }
```
shishirb-MSFT, line 88: >const char* [](http://example.com/codeflow?start=32&length=11)

Without the const char* overload, doesn't the compiler automatically call the std::string& version?

sdk-cpp/include/do_download_property.h
```diff
@@ -23,69 +22,84 @@ class CDownloadPropertyValueInternal;
...
-    static std::error_code make(const status_callback_t& val, download_property_value& out);
 
     std::error_code as(bool& val) const noexcept;
     std::error_code as(uint32_t& val) const noexcept;
     std::error_code as(uint64_t& val) const noexcept;
     std::error_code as(std::string& val) const noexcept;
-    std::error_code as(std::vector<unsigned char>& val) const noexcept;
-    std::error_code as(status_callback_t& val) const noexcept;
+
+#if defined(DO_INTERFACE_COM)
```
shishirb-MSFT, line 98: I'm convinced about this, so we can remove the ifdef.

sdk-cpp/include/do_download_property.h
```diff
@@ -23,69 +22,84 @@ class CDownloadPropertyValueInternal;
...
+    no_surcharge,           // pause download if over data limit
+};
 
-/*
-CDownloadImpl is declared as a friend class because it needs to access the platform-specific native value for download_property_value
-The type of the native value is defined in CDownloadPropertyValueInternal, because DO header files are platform agnostic
-This is so any user of the SDK does not have to worry about supplying platform specific compile definitions to use the SDK
-*/
+// Values for download_property::security_flags
+enum class download_security_flags : uint32_t
```
shishirb-MSFT, line 69: >download_security_flags [](http://example.com/codeflow?start=11&length=23)

Want to talk about this tomorrow.

sdk-cpp/src/internal/do_download_property_internal.h
```diff
@@ -24,22 +23,35 @@ namespace deliveryoptimization
...
 {
 
+#if defined(DO_INTERFACE_COM)
+struct unique_variant : VARIANT
+{
+    unique_variant();
+    explicit unique_variant(const VARIANT& other) noexcept; // takes ownership via shallow copy
+    unique_variant(const unique_variant& other); // true copy
+    unique_variant(unique_variant&& other) noexcept;
+    unique_variant& operator=(unique_variant&& other);
```
shishirb-MSFT, line 33: >operator= [](http://example.com/codeflow?start=20&length=9)

Move assignment must also be noexcept.

sdk-cpp/src/internal/do_download_property_internal.h
```diff
@@ -24,22 +23,35 @@ namespace deliveryoptimization
...
 {
 public:
 #if defined(DO_INTERFACE_COM)
-    using native_type = VARIANT;
+    using native_type = unique_variant;
 #else
-    using native_type = boost::variant<std::string, uint32_t, uint64_t, bool, std::vector<unsigned char>>;
+    using native_type = boost::variant<std::string, uint32_t, uint64_t, bool>;
 #endif
     CDownloadPropertyValueInternal();
```
shishirb-MSFT, line 49: >CDownloadPropertyValueInternal [](http://example.com/codeflow?start=4&length=30)

We probably don't need any of the user-defined c'tor/d'tor now.

sdk-cpp/include/do_download_property.h
JeffreySaathoff: Haven't looked at the code generation, but it picks the wrong overload (probably bool), so we need these overloads.

shishirb-MSFT:
FYI - when completing the PR, github adds all commit messages into the squashed commit. Need to remove/re-format as needed.
Also, add the "SDK:" prefix to the title before merging.. helps in building the release notes in future.

shishirb-MSFT (approved):
:shipit:

PR #156, opened 2023-02-09
Add support for streamed output

JeffreySaathoff:


Jerry-GeGarcia (approved):

sdk-cpp/tests/download_properties_tests.cpp
```diff
@@ -198,6 +198,22 @@ TEST_F(DownloadPropertyTests, ForegroundBackgroundRace)
...
 }
 
+TEST_F(DownloadPropertyTests, BasicStreamTest)
+{
+    auto simpleDownload = msdot::download::make(g_smallFileUrl);
+
+    uint64_t downloadedBytes = 0;
+    simpleDownload->set_output_stream([&downloadedBytes](const unsigned char*, size_t cb) -> std::error_code { downloadedBytes += cb; return DO_OK; });
+
+    simpleDownload->start_and_wait_until_completion();
```
shishirb-MSFT, line 208: >start_and_wait_until_completion [](http://example.com/codeflow?start=20&length=31)

get_property after completion will fail on xplat. Use TestHelpers::WaitForState to wait for transferred and call finalize() at the end.

shishirb-MSFT (approved):
:shipit:

PR #157, opened 2023-02-11
Add support for partial downloads (ranges)

JeffreySaathoff:


sdk-cpp/include/do_download_status.h
```diff
@@ -74,9 +74,17 @@ class download_status
 class download;
 using status_callback_t = std::function<void(download&, download_status&)>;
 
-// Note: std::span might be nice here, but it requires C++20
+// Future: use std::span (requires C++20)
 using output_stream_callback_t = std::function<std::error_code(const unsigned char* data, size_t size)>;
 
+// TODO: create a new header for things that aren't 'status' related, such as output_stream_callback_t and ranges
+struct range
```
shishirb-MSFT, line 81: >range [](http://example.com/codeflow?start=7&length=5)

Call it download_range.

sdk-cpp/src/internal/com/download_impl.cpp
```diff
@@ -263,6 +267,37 @@ std::error_code CDownloadImpl::GetProperty(msdo::download_property key, msdo::do
...
+
+    void *pvBuffer = ::operator new(cbBufferSize, std::nothrow);
+    if (pvBuffer == nullptr)
+    {
+        return make_error_code(E_OUTOFMEMORY);
+    }
+
+    std::unique_ptr<DO_DOWNLOAD_RANGES_INFO> spRanges(new(pvBuffer) DO_DOWNLOAD_RANGES_INFO());
+    spRanges->RangeCount = static_cast<uint32_t>(count);
+    if (count > 0)
```
shishirb-MSFT, line 292: >if (count > 0) [](http://example.com/codeflow?start=4&length=14)

I think we should fail the call at the upper layer itself if count == 0.

shishirb-MSFT (approved):
:shipit:

PR #162, opened 2023-03-10
Enable bootstrap.sh to determine OS and version on its own

shishirb-MSFT:


build/scripts/bootstrap.sh
```diff
@@ -125,82 +111,190 @@ function installBuildDependencies
...
+function determine_machine_architecture()
+{
+    local arch=''
+    arch="$(uname -m)"
+    local ret_val=$?
+    if [[ $ret_val != 0 ]]; then
+        echo "Failed to get cpu architecture."
+        return 1
+    else
+        if [[ $arch == aarch64* || $arch == armv8* ]]; then
```
JeffMill, line 196: see my other comment -- these are mutually exclusive.  e.g. $ARCH_FLAVOR="arm32" would be better IMHO

snapcraft-options/snapcraft-agent.yaml
```diff
@@ -29,7 +29,7 @@ parts:
     plugin: nil
     source: .
     override-build: |
-      ./build/scripts/bootstrap.sh --platform ubuntu2004 --install build
+      ./build/scripts/bootstrap.sh --install build
```
JeffMill, line 32: yay!!!

JeffMill (approved):
looks good overall, just some suggestions added.

Is there a task on ADU to fix up the way we call DO?

build/scripts/bootstrap.sh
```diff
@@ -125,82 +111,190 @@ function installBuildDependencies
...
-    fi
+    installDeveloperTools
+    installContainerTools
+    installQemu
 }
 
 function isSupportedLinux()
 {
-    if [[ "$PLATFORM" == "debian10" || "$PLATFORM" == "ubuntu1804" || "$PLATFORM" == "ubuntu2004" ]];
+    if [[ ("$OS" == "ubuntu" && ($VER == "18.04" || $VER == "20.04"))
```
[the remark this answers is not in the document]
shishirb-MSFT: There was no requirement for other versions and we have not tested this on those other versions.

build/scripts/bootstrap.sh
```diff
@@ -125,82 +111,190 @@ function installBuildDependencies
...
-        installDeveloperTools
-        installContainerTools
-        installQemu
-    fi
+    installDeveloperTools
+    installContainerTools
+    installQemu
 }
 
 function isSupportedLinux()
```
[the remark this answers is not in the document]
shishirb-MSFT: Good point. I'll split it out when there is another script that needs the same functionality.

build/scripts/bootstrap.sh
shishirb-MSFT: Yes, good call.

shishirb-MSFT:
Following up offline.

---
In reply to: [1335530563](https://github.com/microsoft/do-client/pull/162#issuecomment-1335530563) [](http://example.com/codeflow?ancestors=1335530563,1335530563)

JeffreySaathoff (approved):
:shipit:

JeffMill (commented):
awesome, thanks

PR #164, opened 2023-03-20
Add support for setting client certificate

JeffreySaathoff:


shishirb-MSFT:
Worth adding a test case? Even if it is with bogus cert data and expecting set_client_cert to fail.

PR #166, opened 2023-03-22
Replace boost::filesystem usage with std::filesystem

shishirb-MSFT:
Always wanted to move to the C++17 filesystem implementation. What prompted the change now: The FindBoost issue on Ubuntu 20.04 involving symlinking /lib to /usr/lib but /include being non-existent. This change removes public dependency on Boost libs entirely, leaving only the tests using boost::program_options.


shishirb-MSFT:
GCC on Ubuntu 18.04 does not have std::filesystem :(

PR #145, opened 2022-12-20
Workaround 20.04 FindBoost issue

shishirb-MSFT:

PR #167, opened 2023-03-22
A more complete fix for non-existent /include path

shishirb-MSFT:
* A prior PR (#145) fixed the issue when building DO itself. However, consumers of DO still face the issue because DO's cmake config file was not updated. This is now fixed.
* Also enable exec permission for the sdk cleanup install script.


sdk-cpp/build/cmake/deliveryoptimization_sdk-config.cmake
```diff
@@ -1,11 +1,65 @@
...
+        # Cleanup temporary variable
+        set(DO_CMAKE_PREFIX_PATH_SAVE)
+    endif ()
+endmacro()
+
+macro(fixup_do_includes_if_needed)
+    if (CMAKE_SYSTEM_NAME MATCHES Linux)
+        # See explanation in macro override_prefix_path_if_needed, above.
+        # Setting CMAKE_PREFIX_PATH does not work for DO SDK though. So update the INTERFACE_INCLUDE_DIRECTORIES property directly.
+        get_target_property(DOSDK_INC_DIR_PRIV Microsoft::deliveryoptimization INTERFACE_INCLUDE_DIRECTORIES)
```
Jerry-GeGarcia, line 34: >DOSDK_INC_DIR_PRIV [](http://example.com/codeflow?start=28&length=18)

where is this set?

Jerry-GeGarcia (approved):
:shipit:

sdk-cpp/build/cmake/deliveryoptimization_sdk-config.cmake
shishirb-MSFT: That is the output variable for get_target_property method call.

Jerry-GeGarcia: thx, not super familiar with cmake.

PR #169, opened 2023-03-23
Replace boost::filesystem usage with std::filesystem

shishirb-MSFT:
* Always wanted to move to the C++17 filesystem implementation. What prompted the change now: The FindBoost issue on Ubuntu 20.04 involving symlinking /lib to /usr/lib but /include being non-existent.
* This change removes public dependency on Boost libs entirely, leaving only the tests using boost.program_options. So, our cmake config file will no longer have to deal with the paths for boost.
* GCC 8, which is the default at least on Ubuntu 18.04, requires use of std::experimental::filesystem. Code has been changed appropriately to support building with both flavors of filesystem.


JeffreySaathoff (approved):
:shipit:

PR #170, opened 2023-04-28
Adding support for building DO snaps for arm64 architecture

andretoyama-msft:
- Adding support for building both the agent and sdk-tests snaps for arm64.
- Updated documentation for building for arm64.
- Adding network plug to sdk-test run in strict confinement.

### DIFF
--- a/build/build-snaps.sh
+++ b/build/build-snaps.sh
@@ -22,7 +22,12 @@ then
         cp ./snapcraft-options/snapcraft-sdk.yaml ./snap/snapcraft.yaml
     fi
 
-    snapcraft
+    if [ $2 -a $2  == "arm64" ]
+    then
+        snapcraft --target-arch=arm64 --destructive-mode --enable-experimental-target-arch
+    else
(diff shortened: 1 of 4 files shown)


build/build-snaps.sh
```diff
@@ -22,7 +22,12 @@ then
         cp ./snapcraft-options/snapcraft-sdk.yaml ./snap/snapcraft.yaml
     fi
 
-    snapcraft
+    if [ $2 -a $2  == "arm64" ]
```
shishirb-MSFT, line 25: >$2 -a $2 [](http://example.com/codeflow?start=9&length=8)

Why is the $2 -a $2 needed? Doesn't this work? `if [ $2 == "arm64" ]`

andretoyama-msft: it works, but it triggers a warning when checking this if condition if the $2 is null. So the first condition is only to remove the warning.

shishirb-MSFT (approved):
:shipit:

PR #171, opened 2023-05-24
Adjust agent snapcraft yaml before first publish

andretoyama-msft:


shishirb-MSFT:
Update the PR title to "Adjust agent snapcraft yaml before first publish"

shishirb-MSFT (approved):
:shipit:

PR #188, opened 2024-01-10
Simplify release pipeline: Remove usage of release-docker-build-steps.yml

shishirb-MSFT:


andretoyama-msft (approved):
Why is this template being removed? Is it just an improvement or the pipeline is handling those steps different?

shishirb-MSFT:
> Why is this template being removed? Is it just an improvement or the pipeline is handling those steps different?

It was made as a template because it was referenced multiple times in this root yaml. Now, with the usage of `parameters.buildTargets` there is only one reference, so we can simplify by having all the build steps in the root yaml itself.

PR #196, opened 2025-03-06
[Edge] Fix libcpp compatibility issues

Penguinwizzard:
To use `free`, we need to import `<cstdlib>`. This is transitively available here through `<cstdint>` in `do_errors.h` with some libcpp builds, but not all, including some recent ones. For spec compliance, we need to explicitly specify it.


shishirb-MSFT (approved):

shishirb-MSFT:
Thanks for your contribution!

PR #200, opened 2025-06-09
Fix unsafe buffer usage in download_impl.cpp

bhkumar007:
Fix unsafe buffer usage in download_impl.cpp by switching from c-style array to std::array.

This fix ensures that the code can be built with -Wunsafe-buffer-usage plugin from clang.

### DIFF
--- a/sdk-cpp/src/internal/com/download_impl.cpp
+++ b/sdk-cpp/src/internal/com/download_impl.cpp
@@ -3,6 +3,8 @@
 
 #include "download_impl.h"
 
+#include <array>
+
 #include "do_download_property_internal.h"
 #include "do_error_helpers.h"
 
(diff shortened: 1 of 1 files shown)


sdk-cpp/src/internal/com/download_impl.cpp
```diff
@@ -18,17 +20,16 @@ namespace details
 
 static msdo::download_state ConvertFromComState(DODownloadState platformState)
 {
-    static const msdo::download_state c_stateMap[] =
-    {
+    static constexpr auto c_stateMap = std::to_array<msdo::download_state>({
         msdo::download_state::created,      // DODownloadState_Created
```
shishirb-MSFT, line 24: std::array::size and ARRAYSIZE is functionally equivalent. So how does this pass the warning?

bhkumar007: You're right that std::array::size() and ARRAYSIZE are functionally equivalent in terms of returning the number of elements. However, the key reason for switching to std::array::size() is to align with modern C++ practices and to address the Clang static analysis warning.

The Clang plugin warning is typically triggered by the use of raw arrays and macros like ARRAYSIZE, which can be error-prone in certain contexts (e.g., when passed to functions or templates). By switching to std::array, we:

Avoid raw array decay to pointers, which can lead to incorrect size calculations.
Leverage type safety and member functions like .size() that are well-understood by static analysis tools.
Suppress false positives from the Clang plugin, which is designed to flag potential buffer overflows or unsafe array usage patterns.
So while the logic is the same, the change improves code safety, readability, and tooling compatibility.

shishirb-MSFT (approved):

PR #205, opened 2026-01-09
Add build support for Ubuntu-24.04 and Debian-12. Also fix unstable tests.

shishirb-MSFT:


build/scripts/bootstrap.sh
```diff
@@ -167,7 +168,7 @@ function installAll
 
 function isSupportedLinux()
 {
-    if [[ ($DISTRO == "ubuntu" && ($VER == "18.04" || $VER == "20.04" || $VER == "22.04"))
+    if [[ ($DISTRO == "ubuntu" && ($VER == "18.04" || $VER == "20.04" || $VER == "22.04" || $VER == "24.04"))
```
Jerry-GeGarcia, line 171: should 18.04 be removed here?

azure-pipelines/build/linux/du/dopapt-docker.yml
```diff
@@ -46,12 +46,17 @@ parameters:
     - debian11_arm32
     - debian11_arm64
     - debian11_x64
+    - debian12_arm32
+    - debian12_arm64
+    - debian12_x64
     - ubuntu1804_arm64
```
Jerry-GeGarcia, line 52: should 18.04 be removed?

Jerry-GeGarcia (approved):

build/scripts/bootstrap.sh
shishirb-MSFT: Probably. However, our tests are dependent on 18.04 (based on the comment in the root yml files). So, we will have to migrate it to newer versions. Effort vs benefit. Let's talk offline.

